PR #12, opened 2016-02-19
SHAME: Shame on my family for this commit

zachgoins:
There have been so many configuration changes to the boat and I have been a bad git manager that this commit is shameful. BUT, code still need to be updated. 

Included: 
- Update sim to new boat thruster locations and dynamics
- Update thruster mapper to add new thrusters and angles
- Add new controller and path planner
- Update RC node to add kill system and tools to test the controller


jpanikulam:
+1 dishonor

- Jake

PR #13, opened 2016-02-21
GNC: bug fixes at lake day, and PD tuning

jnez71:
Bugs fixed:

- move_helper function had current yaw in radians but accepted the yaw delta value in degrees
- thruster allocator call rate needed to be 50 Hz
- MRAC controller was multiplying gains and errors with * instead of np.dot
- weird murph_sim dependency in simulator didn't need to be there

Also:

- finished MRAC controller documentation
- tuned the PD parts of the MRAC controller at the lake

### DIFF
--- a/gnc/navigator_controller/src/mrac_controller.py
+++ b/gnc/navigator_controller/src/mrac_controller.py
@@ -2,11 +2,82 @@
 '''
 Model-Reference Adaptive Controller
 
-###Purpose
-###How to use
-###Conventions, units, and array orders
-###Explanation of controller
-###References
+################################################# BLACK BOX DESCRIPTION
+
+Inputs:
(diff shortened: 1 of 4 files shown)


gnc/navigator_controller/src/mrac_controller.py
```diff
@@ -22,14 +93,15 @@ class MRAC_Controller:
 
     def __init__(self):
         '''
-        ###Explain what is assigned here
+        Set hard-coded tunable parameters (gains and desired vehicle speed limit).
+        The boat model section's values are used
```
jnez71: Oops, that's embarrassing.. not sure where the rest of that sentence went
I'll touch that up with an inevitable PR at the next lake day.

PR #14, opened 2016-03-31
THRUSTERS and SIM: fix thruster ordering, add effort scaling, fix B-matrix, update sim

jnez71:
There were a ton of places where mismatches in "which thrusters are where" (in a list of four of them) was causing chaos in the sim, and probably in real life. The unified standard is now:

BL, BR, FL, FR
(it's alphabetical, in case you forget)

I also fixed a bunch of other things:

- Thrust values in newtons were being sent straight to the thrusters. I multiplied in a conversion factor, which to my knowledge should map 220 N (measured max thrust) to 600 effort points (presumed max firmware value). This factor is a rosparam called effort_ratio.
- cos() and sin() of thruster angles in degrees were being taken, so I converted the angles to radians first.
- I replaced a manually implemented clip function with np.clip
- I deleted lines of code that did literally nothing
- The y-component of thrust directions was negated before the lever-arm cross-product was taken in the calculation of the B-matrix. Why? Probably fixed things in the past, but it was wrong fixing wrong. The problem was thruster ordering inconsistencies [BL, BR, FL, FR]. Now we have right fixing right.
- I updated the sim to have physical parameters that best match NaviGator (instead of PropaGator)

The sim is actually useful now, but it would be nice if it also visualized the desired waypoint simultaneously. I may do that another day.

### DIFF
--- a/gnc/navigator_thrust_mapper/src/thrust_mapper.py
+++ b/gnc/navigator_thrust_mapper/src/thrust_mapper.py
@@ -32,6 +32,8 @@ BR = Back Right
 FL = Front Left
 FR = Front Right
 
+^ in that order (top to bottom), it's also alphabetical if you forget
+
 Bibliography:
     [1] Christiaan De Wit
         "Optimal Thrust Allocation Methods for Dynamic Positioning of Ships"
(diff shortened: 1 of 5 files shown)


gnc/navigator_thrust_mapper/src/thrust_mapper.py
```diff
@@ -84,87 +91,82 @@ def wrench_cb(self, msg):
...
-        for thruster, position in enumerate(self.positions):
+        for thruster_number, position in enumerate(self.positions):
             # l_x and l_y are the offset for the center of gravity
             l_x, l_y = np.subtract(position, self.boat_cog)
             # sin and cos of the angle of the thrusters
-            cos = np.cos(angles[count])
-            sin = np.sin(angles[count])
-            torque_effect = np.cross((l_x, l_y), (cos, -sin))
+            cos = np.cos(self.angles[thruster_number])
+            sin = np.sin(self.angles[thruster_number])
```
zachgoins: Goooood this is the way I was intending on changing it to. Excellent

zachgoins:
Looks like there is something off with the sim thrust vector visualization. Going to merge anyways because everything else looks good and that is an insignificant things to worry about for now.

PR #17, opened 2016-04-25
Lake Day 4/23/16

jnez71:
Everything worked at the lake, and I also have it all working in simulation too.
- new NN controller option (if you want to use it, update the dependency list)
- final (probably) tuning of MRAC controller
- new MRAC feature: subscribes to a Bool /mrac_learn to start and stop learning (defaults to stopped)
- option to use C3 trajectory generator from sub instead of the built-in one in MRAC (the NN_controller, on the other hand, must use an external trajectory generator like C3; there is none built in).

I left the system set up to use MRAC with internal trajectory generator, as it is compatible with Zach's gui and is a lot more predictable.

### DIFF
--- a/gnc/navigator_controller/src/mrac_controller.py
+++ b/gnc/navigator_controller/src/mrac_controller.py
@@ -122,8 +122,8 @@ class MRAC_Controller:
         # Limits on drag estimate effort
         self.drag_limit = np.array([1000, 1000, 1000])
         # User-imposed speed limit
-        self.vel_max_body_positive = np.array([1.5, 0.75, 0.5])  # [m/s, m/s, rad/s]
-        self.vel_max_body_negative = np.array([0.75, 0.75, 0.5])  # [m/s, m/s, rad/s]
+        self.vel_max_body_positive = np.array([1.1,  0.45, 0.19])  # [m/s, m/s, rad/s]
+        self.vel_max_body_negative = np.array([0.68, 0.45, 0.19])  # [m/s, m/s, rad/s]
         # "Smart heading" threshold
         self.heading_threshold = 500  # m
         # Only use PD controller
@@ -134,8 +134,8 @@ class MRAC_Controller:
 
         #### REFERENCE MODEL (note that this is not the adaptively estimated TRUE model; rather,
         #                     these parameters will govern the trajectory we want to achieve).
-        self.mass_ref = 300  # kg
-        self.inertia_ref = 300  # kg*m^2
+        self.mass_ref = 400  # kg, determined to be larger than boat in practice due to water mass
+        self.inertia_ref = 400  # kg*m^2, determined to be larger than boat in practice due to water mass
         self.thrust_max = 220  # N
         self.thruster_positions = np.array([[-1.9000,  1.0000, -0.0123],
                                             [-1.9000, -1.0000, -0.0123],
--- a/gnc/navigator_controller/src/mrac_controller.py
+++ b/gnc/navigator_controller/src/mrac_controller.py
@@ -170,7 +170,7 @@ class MRAC_Controller:
         self.ang_vel = 0
         self.state = Odometry()
         # Subscribers
-        rospy.Subscriber("/waypoint", Point, self.set_waypoint)
+        rospy.Subscriber("/waypoint", PoseStamped, self.set_waypoint)
         rospy.Subscriber("/odom", Odometry, self.get_command)
         # Publishers
         self.wrench_pub = rospy.Publisher("/wrench/autonomous", WrenchStamped, queue_size=0)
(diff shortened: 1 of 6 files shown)


LucasBA:
"Everything worked at the lake" I know what all of these words mean individually but together, and in this order the phrase sounds nonsensical.

forrestv:
>I left the system set up to use MRAC with internal trajectory generator, as it is compatible with Zach's gui and is a lot more predictable.

You forgot "untuned" and "unvalidated"...

We spent a while setting up the acceleration and velocity limits for c3 at the lake, and only then was the boat somewhat "working," as I remember. Moving to c3 is annoying, but MRAC's trajectory generator was neither satisfactorily tuned nor tested, so why make MRAC's the default?

It might not affect the code much (it's only "self.use_external_tgen = False" that's relevant, I think), but I'm confused by your summary.

jnez71:
I'll change its top speed and accel parameters to match what we tuned for C3. I have validated that it never exceeds these constraints, even when doing coupled motions. Additionally, when doing uncoupled motions, they are just as predictable as C3; always straight paths. I think you saw it at a weird time in its development, or while I was doing coupled rotations and translations (motion that c3 can't even attempt) Feel free to play with it in sim.

jnez71:
I have no problem with using C3 instead, if Zach or whoever is using the boat this summer would prefer it. Like you said, just change self.use_external_tgen to True in mrac_controller.py. I really just wanted to leave Zach with something that works with his GUI.

zachgoins:
I haven't been able to look through everything yet but I will check it all
out tomorrow. I would much rather use C3 though since it works with pose
editor, but if I can change it use C3 then we should be good to go. The
priority is solving the boat, so if the best t gen is C3 then I will change
whatever I need to change to make that work.

On Monday, April 25, 2016, Jason Nezvadovitz <notifications@github.com>
wrote:

> I have no problem with using C3 instead, if Zach or whoever is using the
> boat this summer would prefer it. Like you said, just change
> self.use_external_tgen to True in mrac_controller.py. I really just wanted
> to leave Zach with something that works with his GUI.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/uf-mil/Navigator/pull/17#issuecomment-214579149>
>


-- 
*Zach Goins | 918.801.8242*
University of Florida, '16
Business | Computer Science and Engineering
Github <https://github.com/zachgoins> | LinkedIn
<https://www.linkedin.com/in/zachgoins>

jnez71:
Oh, okay then yeah I'll change the default to C3 so you can use pose editor. Remember to do
`rosrun kill_handling clear`
once things start up for C3 to actually begin working (unless you are actually using kill_handling for something else, in which case just do what you do to unkill).

jnez71:
So I just fixed a couple bugs..

When MRAC switched to using the external tgen, it didn't convert the external tgen's twist and accel into world frame before doing world frame math. (In general MRAC is kinda naively done - babby's first controller - I should have done math in body frame, it's much simpler, but since the bug is fixed lets just roll with it). That may have incorrectly influenced our tuning at the last lake day because the acceleration feedforward term was actually fighting us sometimes. At the lake just pay attention to if the thruster commands are saturating and lower the kp and kd gains together evenly until thrusters don't saturate. Of course contact me if you have any questions.

The simulator would initialize the orientation state to [0, 0, 0, 0] which caused an inf warning when you open it, but just one because then it fixes itself. That warning bugged me, so I prevented the [0, 0, 0, 0] situation even if just for an instant.

gnc/navigator_controller/src/mrac_controller.py
```diff
@@ -336,7 +377,12 @@ def increment_reference(self):
 
         # Use model drag to find drag force on virtual boat
         twist_body = R_ref.T.dot(np.concatenate((self.v_ref, [self.w_ref])))
-        D_body = np.array([p if v > 0 else n for p, n, v in zip(self.D_body_positive, self.D_body_negative, twist_body)])
```
zachgoins: This is crafty

gnc/navigator_controller/src/mrac_controller.py
```diff
@@ -336,7 +377,12 @@ def increment_reference(self):
...
+        D_body = np.zeros_like(twist_body)
```
zachgoins: Doesn't this make the line above it unnecessary?

zachgoins:
I am going to merge because I want to work on this stuff in another branch and the line note I added is trivial. Looks good!

gnc/navigator_controller/src/mrac_controller.py
jnez71: zeros_like() is the same as zero(len())
That line initializes D_body, it does not do anything to twist_body except reference its len().

gnc/navigator_controller/src/mrac_controller.py
jnez71: I changed its structure because it always left D_body with size (1,) for some reason instead of filling it properly like the normal for loop I replaced it with does

PR #20, opened 2016-07-07
Added navigator_msgs dependency to navigator_lidar

ironmig:
When building Navigator from a fresh install after following the wiki tutorial, I encountered this error:
`
/WORKSPACE/src/NaviGator/sensor_control/navigator_lidar/src/cluster_extraction.cpp:6:34: fatal error: navigator_msgs/Point.h: No such file or directory
 #include "navigator_msgs/Point.h"
                                  ^
compilation terminated.
` 

I don't really understand ROS or the whole build process of this project yet, but after some digging around on google I found that adding the project that generates Point.h to the find_package thing in the lidar cmake file made the build work.


jpanikulam:
Nice noodling, Kevin.

PR #25, opened 2016-08-09
INFRASTRUCTURE: Install Script and Structure Changes

whispercoros:
The install script was created due to popular demand from the software team. It is just based on the install instructions on the wiki. I recommend that we look into cleaning up the overall structure of the repository, which should be discussed at the software meeting. Building with catkin_make was successful. On a side note, I am looking into setting up Semaphore CI for the Navigator repository.


install.sh
```diff
@@ -0,0 +1,279 @@
...
+	cd $CATKIN_DIR/src/Navigator
+	git remote rename origin upstream
+	instlog "Make sure you change your git to point to your own fork! (git remote add origin your_forks_url)"
+fi
+
+# Check if the Sub8 repository is present; if it isn't, download it
+if !(ls "$CATKIN_DIR/src" | grep --quiet "Sub8"); then
+	instlog "Downloading up the Sub8 repository"
+	cd $CATKIN_DIR/src
+	git clone -q https://github.com/whispercoros/Sub8.git
```
ironmig: Should this be to https://github.com/uf-mil/Sub8.git?

install.sh
```diff
@@ -0,0 +1,279 @@
...
+source "$CATKIN_DIR/devel/setup.bash"
+if !(cat ~/.bashrc | grep --quiet "source $CATKIN_DIR/devel/setup.bash"); then
+	echo "source $CATKIN_DIR/devel/setup.bash" >> ~/.bashrc
+fi
+
+# Check if the Navigator repository is present; if it isn't, download it
+if !(ls "$CATKIN_DIR/src" | grep --quiet "Navigator"); then
+	instlog "Downloading up the Navigator repository"
+	cd $CATKIN_DIR/src
+	git clone -q https://github.com/whispercoros/Navigator.git
```
ironmig: Should this be to https://github.com/uf-mil/Navigator.git?

whispercoros:
Good catch; that was a bit of debugging code I must have missed before pushing. It is fixed now.

whispercoros:
I have done some in-depth testing and fixed some small edge-case issues. The state that the script is in now should work for everyone.

mattlangford:
For the record, I had asked Anthony to move the python files from src/ to nodes/

install.sh
```diff
@@ -0,0 +1,276 @@
...
+
+#===================================#
+# Navigator Dependency Installation #
+#===================================#
+
+instlog "Installing Navigator dependencies from the Ubuntu repositories"
+sudo apt-get install -qq libompl-dev
+sudo apt-get install -qq python-pygame
+
+instlog "Installing Navigator ROS dependencies"
```
mattlangford: Might need [this](http://wiki.ros.org/compressed_image_transport)? @rleyva

rleyva: Yes, please.

mattlangford: Might put it under Sub dependancies - I don't know

install.sh
```diff
@@ -0,0 +1,276 @@
...
+sudo pip install -q -U tqdm
+sudo pip install -q -U service_identity
+sudo pip install -q -U pyasn1
+sudo pip install -q -U characteristic
+sudo pip install -q -U crc16
+
+instlog "Getting Sub8 ROS packages we need to install from source"
+ros_git_get https://github.com/uf-mil/Sub8.git
+ros_git_get https://github.com/txros/txros.git
+ros_git_get https://github.com/uf-mil/rawgps-tools.git
```
mattlangford: Does this install the most recent rawgps-tools? It has the broken commit in master still.

install.sh
```diff
@@ -0,0 +1,276 @@
...
+sudo pip install -q -U argcomplete
+sudo pip install -q -U scikit-learn > /dev/null 2>&1
+sudo pip install -q -U tqdm
+sudo pip install -q -U service_identity
+sudo pip install -q -U pyasn1
+sudo pip install -q -U characteristic
+sudo pip install -q -U crc16
+
+instlog "Getting Sub8 ROS packages we need to install from source"
+ros_git_get https://github.com/uf-mil/Sub8.git
```
mattlangford: Do we want this?

install.sh
rleyva: The most recent rawgps broke the sub because it didn't have GPS, it shouldn't affect the boat. (I think it was GPS....)

install.sh
rleyva: Ideally there should be two install scripts, or an environment variable to determine what type of machine we're on. For semaphore ros-indigo-desktop is fine, but everyone should be running ros-indigo-desktop-full on their machine. This would remove our issues with having to enumerate every single package. Do we have any idea if the time hit with installing desktop-full on semaphore is big enough to make desktop more appealing?

install.sh
rleyva: mayavi was used for visualizing on some sub thresholding tool, I think so that one is not necessary. The other ones I'd say are useful to have except maybe crc16.

whispercoros: According to the Navigator wiki, we need the Sub8 repository in order for this one to work properly. As much as I would like that to change, it would probably be too much of a hassle. Can mayavi be removed safely (e.g. catkin_make will still compile Sub8 and it will be usable)?

install.sh
whispercoros: Is there a unit test we can run to determine if this will be an issue, or will we just have to see on testing day? Either way, we can always revert the repository manually if need be and that step can be absorbed into the script later on.

install.sh
whispercoros: I already install all the compression tools I found useful on line 245:

`sudo apt-get install -qq ros-indigo-rosbag-image-compressor ros-indigo-compressed-image-transport ros-indigo-compressed-depth-image-transport`

I will add a commit to change ros-indigo-desktop to ros-indigo-desktop-full for now. I have Semaphore slated for after I move tomorrow, so the script will likely see some changes after that gets set up. As for the time hit, it takes me 2 minutes to add an environment variable and an if statement, so it is justifiable.

whispercoros:
My machine is complaining about missing ros-indigo-desktop-full dependencies that cannot be installed. This may be a side effect only on my machine, but I need to do some sleuthing to be sure. Do not merge this until I fix this.

install.sh
mattlangford: I'd rather keep mayavi, it's a better alternative to Matplotlib

whispercoros:
There is a conflict with ros-indigo-desktop-full and gazebo7 (I believe the former installs Gazebo2). I'll fix this later tonight by breaking the metapackage.

install.sh
jpanikulam: <3

- Jake

whispercoros:
I have just applied the fix and tested it, so this can now be merged in.

PR #26, opened 2016-08-11
INFRASTRUCTURE: Add script for configuring Navigator networking

ironmig:
Since we all seem to be in a scripting spirit:
Adds a bash script which does everything described in [Networking with Navigator](https://github.com/uf-mil/Navigator/wiki/Networking-with-NaviGator) and makes it easy/automated to connect and disconnect from navigator.

to use:
```
sudo ./network-config -i -d eth0 -a 192.168.1.200
```
This automatically configures your .bashrc, /etc/hosts, and network-manager config to use with Navigator. After running this (and sourcing .bashrc again) you can connect with ```navigator-connect```, which sets the correct ROS_MASTER_URI and tells network-manger to connect to the Navigator ethernet profile. ```navigator-disconnect``` does the opposite.

I'll continue to work on it but wanted to put the PR up for comments.

Current issues:
- Doesn't play well with computers already set up to connect with Navigator, best to clear the Navigator parts of /etc/hosts and .bashrc before running
- No way to change static IP or eth interface after installing without manually opening files


whispercoros:
We are currently working on switching the network to something more robust than just setting static IP addresses. This will eliminate the need to configure network manager, since IP's will be managed through DHCP. The hosts file will be managed by a script that was created for Sub8, which will allow us to keep it in sync with the DHCP static host reservations. When I am done working on the rest of the install script, I will merge the aliases that you created for the bashrc file into the install script.

whispercoros:
On a side note, this script will be useful for anyone who has not already configured their interfaces by Saturday. Anyone who wants to use it should run the following:

    wget -O https://raw.githubusercontent.com/ironmig/Navigator/7a7fa3e9230fa522c30229c465b628692a2e1a77/network-config network-config.sh && chmod +x network-config.sh

ironmig:
Added to #49, closing

PR #30, opened 2016-08-18
Alarms, Launch, Utils

mattlangford:
Moved utils over from sub repo to remove the dependency there (I think Jake wrote most of those utils).

The alarms should work now - not sure if they were working before but from what I saw they shouldn't have been to the full extent. There is a `kill` alarm and a `docking` alarm to replace `full_kill` and `half_kill` alarms. `kill` will shutdown C3 as well as zeroing motor commands.

Cameras will need to be calibrated


mission_control/navigator_launch/launch/cameras.launch
```diff
@@ -0,0 +1,80 @@
+<launch>
+  <!-- This is the default pointgrey launch file with our id's and our names  -->
+
+  <arg name="front_camera_serial" default="15223362" />
```
DSsoto: What are the serial parameters for? Do they identify each camera?

mission_control/navigator_launch/launch/cameras.launch
```diff
@@ -0,0 +1,80 @@
...
+  <arg name="front_camera_calibrated" default="0" />
+  <arg name="down_camera_serial" default="10282860" />
+  <arg name="down_camera_calibrated" default="0" />
```
DSsoto: Isn't it more readable to use booleans?

rleyva: Werid... These are PointGrey launches, not dc1394 ones. I'm not entirely sure why these are still around. These are not the launches that are running the cameras, and if this is what's being used to launch we need to switch over to the launches for dc1394.

gnc/navigator_thrust_mapper/nodes/thrust_mapper.py
```diff
@@ -172,7 +162,6 @@ def allocate(self):
             self.FR_pub.publish(FR_msg)
 
 
-
 if __name__ == "__main__":
 
     rospy.init_node('primitive_driver')
```
DSsoto: This is not a good name for such a crucial node. Why not "thrust(er)_mapper"?

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
```diff
@@ -5,53 +5,56 @@
...
 
         self.force_scale = rospy.get_param("~force_scale")
         self.torque_scale = rospy.get_param("~torque_scale")
 
-        self.last_controller_state = 0;
-        self.last_station_hold_state = 0;
-        self.last_kill_state = 0
-        self.last_revive_state = 0
-        self.current_pose = Odometry()
+        self.last_change_mode = 0
```
DSsoto: I don't think we should use integers in place of boolean values just because they can be casted.

mission_control/navigator_launch/launch/tf.launch
```diff
@@ -0,0 +1,6 @@
+<node pkg="tf" type="static_transform_publisher" name="ins_tf" args=".37084 .0.05715 .7493  0 0 0 1  /base_link /ins 100"/>
+<node pkg="tf" type="static_transform_publisher" name="lidar_tf" args=".5334 0.0127 .7493  0 0 0 1  /base_link /lidar 100" />
+<node pkg="tf" type="static_transform_publisher" name="forward_camera_tf" args=".6096 -.254 .762  0 0 0 1  /base_link /camera 100" /> 
+<!--node pkg="tf" type="static_transform_publisher" name="right_camera_tf" args=".6096 -.254 .762  0 0 0 1  /base_link /camera 100" /--> 
+<!--node pkg="tf" type="static_transform_publisher" name="down_camera_tf" args=".6096 -.254 .762  0 0 0 1  /base_link /camera 100" /--> 
+<!--node pkg="tf" type="static_transform_publisher" name="shooter_tf" args=".6096 -.254 .762  0 0 0 1  /base_link /camera 100" /-->
```
DSsoto: Dead code, you should add a comment saying these are very rough approximations.

mission_control/navigator_launch/launch/cameras.launch
DSsoto: You're the optometrist, your call. They do seem to be working ok. What was the main reason we decided to switch to 1394?

rleyva: We'll get rid of them once I get to Gainesville and verify there's not something screwy happening with the cameras. I set up the cameras/launches forever ago so I might have forgotten something, or someone might have changed something.

rleyva: Also, the Point Grey drivers absolutely suck, I spent months trying to run more than one camera at the same time and it never worked. The new drivers they put out might work better but I don't care to find out. The sub runs 1394, and last I checked the boat was also running 1394.

mission_control/navigator_launch/launch/cameras.launch
mattlangford: Yep

gnc/navigator_thrust_mapper/nodes/thrust_mapper.py
mattlangford: Will it be changed with Jason's purge?

DSsoto: I'm not sure if he was planning on it, but it should be.

mission_control/navigator_launch/launch/cameras.launch
mattlangford: There was a camera launch on the boat that would launch just one camera but the image was black and white in rqt. This launch file launches all three cameras with the processing. I just wanted to get it working to bag on Saturday. If you'd like to scrap it afterwards, i won't get in your way

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
mattlangford: The joystick returns ints as the button states so it's easier to compare and set that way

mission_control/navigator_launch/launch/cameras.launch
rleyva: Go for it, do whatever you have to do to get it up and running for Saturday. Just make sure that PTGrey launch works.

mattlangford:
Good to merge?

mattlangford:
Merging... If you got beef, let me know and I will attempt to correct the issues

Issue #24, opened 2016-08-06
Turn Shooter into ROS Service


zachgoins:
We need a way to be able to fire the launcher from other programs using something like

shooter.fire(), shooter.cancel(), etc...

Basically just opens up the commands you would manually input into the terminal to the ROS server system where we can call the functions from any programs. 

http://wiki.ros.org/rospy/Overview/Services

LucasBA:
Should this be assigned to @ironmig ?

rleyva:
I added him to uf-mil. Once he accepts the invite we'll assign him to the issue.

ironmig:
Any suggestions on how to lay this out? Is it better to have a unique service for each operation (shoot, cancel) or just one service where you pass data about what operation to do?

rleyva:
Take a look at sub8_actuator_driver, the service model there should port over directly. Here is the [link](https://github.com/uf-mil/Sub8/blob/5feeed1af81630d056c772f909712621ea6f0c77/drivers/sub8_actuator_driver/srv/SetValve.srv) to the service. Take a look at the readme in the package for more info.

zachgoins:
The link that Ralph gave is great, but yes typically there is a different
service for each action. They can all be in the same server file but we
want to be able to call your command line inputs directly ie
shooter.shoot(), shooter.cancel(), and then the server interacts with the
shooter script appropriately.

*Zach Goins | 918.801.8242*
paracosm.io
Github <https://github.com/zachgoins> | LinkedIn
<https://www.linkedin.com/in/zachgoins>

PR #32, opened 2016-08-23
SHOOTER: Control from ros services

ironmig:
As per #24, the shooter can be controlled with a service.

* A std_srvs/Trigger call to /shooter/fire will fire balls (currently all 4, will change with new loader)

* A std_srvs/Trigger call to /shooter/cancel will stop any movement on the shooter

* A custom service can manually control the two motors on the shooter for testing

* The arduino ros client has also been added to the motor controller launch file so it gets started when the boat starts up
* Launch file for arduino client has been updated to use the correct USB port ID


ironmig:
I'll look into the CI error. Perhaps it didn't rebuild the CMAKE stuff to see the new service file? By the way this code was tested on the lake on 8/20 and worked as expected.

DSsoto:
Not sure what it is, but it's not that. Semaphore downloads the repo and builds from scratch every time.

mattlangford:
Not sure if you have access to see where the build failed but this is the message:
```
/home/runner/navigator_ws/src/Navigator/mission_systems/shooter/firmware/shooter.cpp:5:42: fatal error: navigator_msgs/ShooterManual.h: No such file or directory
 #include <navigator_msgs/ShooterManual.h>
```

mattlangford:
After lake day I should have some free time that we can sit down and diagnose the issues here
I don't want to merge yet, since any subsequent builds would fail too

mattlangford:
Semaphore looks like it builds correctly now (though it won't with an old `install.sh`). If you guys want to bring this up to the latests master, we will merge

mission_control/navigator_launch/launch/motor_controller.launch
```diff
@@ -15,6 +15,9 @@
     <param name="port" value="/dev/serial/by-id/usb-FTDI_Roboteq_MDC1460_stbd-bow-if00-port0"/>
     <param name="value" value="115200" />
   </node>
+  <node pkg="rosserial_python" type="serial_node.py" name="arduino_serial">
```
mattlangford: Can you make a new `hw_drivers.launch` and put this in there? We will put the hydrophones launch in there and any other hardware interfaces (excluding the motor drivers).

mission_control/navigator_launch/launch/master_boat.launch
```diff
@@ -1,5 +1,6 @@
 <launch>
     <include file="$(find navigator_launch)/launch/motor_controller.launch"/>
+    <include file="$(find navigator_launch)/launch/hw_drivers.launch"/>
```
mattlangford: Good catch

Issue #7, opened 2015-10-02
Transition Heartbeat and Communication Monitoring


zachgoins:
Transition the heartbeat monitoring used on PropaGator 2

whispercoros:
Since this was fixed in #35, can it be closed?

PR #35, opened 2016-08-25
Fix Thruster Issues, Network Kills

### LINKED ISSUE #34: Thrusters config reset
The thrusters config get reset whenever the driver gets launched. This causes problems with the controller since the default acceleration is low.

mattlangford:
I think this addresses #7, what do you guys think?
Also fixes #34 

We still require the sub repo as a dependency for just sub8_alarms I believe - we should find a way to move that package whilst keeping the git history intact. [It looks possible](http://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history), albeit dangerous and cooky.


hardware_drivers/roboteq/roboteq_driver/mbs/script.hex
```diff
@@ -1,38 +1,38 @@
...
-:1001C0000204052A2524013A02020204052A2524F4
-:1001D000013A02030204052A2524013A020402041A
-:1001E000052A2524013A020D02022A2524013A0299
-:1001F0000C0204052A2524013A02100204052A25CE
-:1002000024013A02120204050201092A2524010DE3
-:1002100022240326733A01052524013A0215012AF6
-:100220002524013A0214012A2524013A020D020173
-:100230002A2524013A020D02032A2524013A02123A
-:1002400002012A2524010D2200FFFFFFFFFFFFFF0F
+:1000300002020127020201032C012702030102B47C
```
DSsoto: Where did you get this from? What does it do?

mattlangford: It's the .mbs script compiled into hex, as per their instructions on [this readme](https://github.com/uf-mil/Navigator/blob/master/hardware_drivers/roboteq/roboteq_driver/mbs/README.md). It's a configuration flashed to each thruster on startup

hardware_drivers/roboteq/roboteq_driver/mbs/script.mbs
```diff
@@ -17,7 +17,7 @@ setcommand(_VAR, 1, -1)
 ' www.roboteq.com/index.php/docman/motor-controllers-documents-and-files/documentation/user-manual/7-nextgen-controllers-user-manual/file
```
DSsoto: Can you delete this file since its not being used?

mission_control/navigator_launch/nodes/network_keepalive.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+import rospy
+from std_msgs.msg import String
+
+
+class KeepAlive(object):
+    '''
+    Meant to run from any number of remote computers.
+    Make sure the wifi link is somewhere between the computers running the keepalive
+        script and the boat (or else there is no point!)
```
DSsoto: Can you add "run on shore, not on the boat", or something along those lines, to make it abundantly clear to new people?

mission_control/navigator_launch/nodes/network_keepalive.py
```diff
@@ -0,0 +1,23 @@
...
+        self.pub = rospy.Publisher('/keep_alive', String, queue_size=1)
+        self.timer = rospy.Timer(rospy.Duration(0.1), self.keepalive_pub)
+
+    def keepalive_pub(self, *args):
+        self.pub.publish(String('keep_alive'))
+
+
+if __name__ == '__main__':
+    rospy.init_node('network_keepalive')
+    KeepAlive()
```
DSsoto: Why aren't you assigning the KeepAlive object to a variable? Won't this object die before the next line?

mission_control/navigator_launch/nodes/network_keepalive.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+import rospy
+from std_msgs.msg import String
+
+
+class KeepAlive(object):
```
DSsoto: For the KeepAlive system, there is a publisher and a listener. Why not make that explicit with the naming of these classes and their respective files? Maybe KeepAlivePublisher and KeepAliveListener? ... KeepAliveServer and KeepAliveClient?

mission_control/navigator_launch/nodes/network_keepalive.py
```diff
@@ -0,0 +1,23 @@
...
+    '''
+    Meant to run from any number of remote computers.
```
DSsoto: "Any number"? Wouldn't it defeat the whole point of this if you run the KA pub on more than one computer?

mission_control/navigator_launch/nodes/network_listener.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+import rospy
+from std_msgs.msg import String
+from sub8_alarm import single_alarm
+from twisted.internet import reactor
+import time
+
+
+class NetworkCheck(object):
```
DSsoto: See naming comment above, I think It would be nice to make it clear that the KeepAlive system is one coherent system with two components just by looking at the names.

mission_control/navigator_launch/nodes/network_listener.py
```diff
@@ -0,0 +1,37 @@
...
+    '''
+    def __init__(self, timeout=2.0):
+        self.timeout = rospy.Duration(timeout)
+        self.sub = rospy.Subscriber('/keep_alive', String, self.got_network_msg, queue_size=1)
+
+        self.alarm_broadcaster, self.alarm = single_alarm('kill', severity=3, problem_description="Network loss")
+        rospy.Timer(rospy.Duration(0.1), self.check)
+
+    def check(self, *args):
+        if self.need_kill() and self.last_msg != '':\
```
DSsoto: Won't self.last_msg always == 'keep_alive' after the first call to got_network_msg?
What is this backslash for?

mission_control/navigator_missions/nodes/run_mission
```diff
@@ -25,15 +25,15 @@ def main():
         print
         defer.returnValue(reactor.stop())
 
-    n = Navigator(nh)._init()
+    n = yield Navigator(nh)._init()
```
DSsoto: What does the change to this line and line 36 do?

mission_control/navigator_launch/nodes/network_keepalive.py
mattlangford: Nope, the class exists in the nether until `rospy.spin()` dies. Want me to switch it back to `all_hail_satan`?

mission_control/navigator_launch/nodes/network_keepalive.py
mattlangford: Well the idea is that we are testing to make sure the network link is up. If only one computer publishes it, then if that computer freezes or something, that doesn't mean the network connection dropped. Since were using tmux, if the one controller computer freezes, someone else could just attach to their session and resume control

mission_control/navigator_launch/launch/gnc.launch
```diff
@@ -44,30 +40,25 @@
...
+    <node pkg="navigator_launch" type="network_listener.py" name="network_kill"/>
+
     <node name="wrench_arbiter" pkg="navigator_msg_multiplexer" type="wrench_arbiter.py"/>
 
     <node name="pose_controller" pkg="navigator_controller" type="mrac_controller.py"/>
     <!-- can be type="mrac_controller.py" OR type="run_nn_controller" if you have that dependency
     <https://github.com/jnez71/neural_control>-->
 
-    <node if="$(arg rc)" pkg="navigator_joystick_control" type="navigator_joystick.py" name="joystick_wrench" output="screen">
+    <node pkg="navigator_joystick_control" type="navigator_joystick.py" name="joystick_wrench" output="screen">
```
DSsoto: Maybe for a future PR it would be nice to have these parameters be dynamic_reconfigure(able) as torque and force sensitivities. Haha it would be really fun to have the bumpers temporarily improve sensitivity, BOOST!

mission_control/navigator_launch/nodes/network_listener.py
mattlangford: NetworkCheck would check for the network messages for the sub. There were two messages for the sub: `auto` and `kill`. A system like that may be set up for the boat in the future.

mission_control/navigator_launch/launch/master_boat.launch
```diff
@@ -1,3 +1,5 @@
 <launch>
```
DSsoto: I don't see why these launches are even called master, there are no slave counterparts.

mission_control/navigator_missions/nodes/run_mission
mattlangford: If you put two missions in the command line, it would have tried to yield on the already yieled `navigator` object when the first mission completed.
Yieleding on a non `deffered` type object will throw errors

mission_control/navigator_launch/launch/master_boat.launch
mattlangford: Master in terms of launching everything - like this is the master launch file for all sub-components

mission_control/navigator_launch/launch/master_shore.launch
```diff
@@ -1,3 +1,9 @@
 <launch>
```
DSsoto: I like what we were doing for the sub where we would extract all of these little params to yaml config files all in the same launch/config directory. What do you think?

mattlangford: Not sure what you mean by all the little params

mission_control/navigator_launch/nodes/network_keepalive.py
DSsoto: Really? I'm surprised, I was trying to find out how often the python garbage collector looked for unreferenced objects but didnt find the answer.

Yes, that would be much better. All hail the dark one.

mission_control/navigator_launch/nodes/network_keepalive.py
DSsoto: Ok, I see.

mission_control/navigator_launch/launch/master_shore.launch
DSsoto: Like the ones one lines 3-5. 

For example, Sub8's perception [launch](Sub8/command/sub8_launch/launch/subsystems/perception.launch), which has tons of parameters, looks very clean.

mattlangford: Wouldn't we just be adding files at that point, there's only 3 params that will (probably) never change. For the perception it makes sense because there are generally a lot of params that will change on the fly.

PR #41, opened 2016-08-29
PERCEPTION: Add initial dock shape vision service and shooter mission

ironmig:
Thought I'd go ahead and start this PR up for Kevin and Daniel's dock symbol vision code and shooter mission. The code is a little ugly and we'll be working the next week or two to refactor, document, and tune it. Just putting it here now for discussion and tracking progress.


mattlangford:
That error looks like it's not a semaphore build error.

RustyBamboo:
It's not generating the header service/message files

sensor_control/vision/symbols/src/smart.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "FrameProc.h"
+#include "ShapeFind.h"
+
+#include <vector>
+
+#include "DebugWindow.h"
+#include "opencv2/core/core.hpp"
+#include "opencv2/highgui/highgui.hpp"
+#include "opencv2/imgproc/imgproc.hpp"
+#include "opencv2/opencv.hpp"
```
DSsoto: The previous 3 are redundant if you use this include.

sensor_control/vision/symbols/src/FrameProc.h
```diff
@@ -0,0 +1,51 @@
...
+#include "DebugWindow.h"
+
+using namespace cv;
+
+class FrameProc
+{
+	private:
+    //const int height = 480;
+		//const int width = 640;
+		static const int blur_kernel_size;
```
DSsoto: Why are you making these static constants? It seems like you should be able to adjust these if you want to use this same object to process images with different parameters.

sensor_control/vision/symbols/src/FrameProc.h
```diff
@@ -0,0 +1,51 @@
+//Grabs frame from ROS, erode/dilate, converts to HSV, produces Red Green Blue Black binary frames;
+#ifndef FROM_PROC_H
```
DSsoto: `#pragma once` does the same thing as these include guards but with much less typing / worrying about unique names.

sensor_control/vision/symbols/src/FrameProc.cpp
```diff
@@ -0,0 +1,85 @@
+#include "FrameProc.h"
+
+
+const int FrameProc::blur_kernel_size = 3;
```
DSsoto: Why don't you just pass these as arguments to the constructor instead?

sensor_control/vision/symbols/src/FrameProc.cpp
```diff
@@ -0,0 +1,85 @@
...
+	green = ColorThresh{Scalar(30, 85, 25),Scalar(85,255,255)}; // 70/2 to 170/2
+	//Green now works... But the crests in the shapes are the same shades as the grass...
+	
+	//http://graphicdesign.stackexchange.com/questions/16166/what-is-the-relationship-between-hue-saturation-and-value
+	
+	rgb_frame = Mat();
+	hsv_frame = Mat();
+	binary_blue_frame = Mat();
+	binary_red_frame = Mat();
+	binary_green_frame = Mat();
```
DSsoto: You could delete these previous 5 lines and nothing would change.

sensor_control/vision/symbols/src/FrameProc.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+	erode_element = getStructuringElement(MORPH_RECT,Size(2*erode_kernel_size + 1,2*erode_kernel_size+1), Point(erode_kernel_size,erode_kernel_size)) ;
+
+	dilate_element = getStructuringElement(MORPH_RECT,Size( 2* dilate_kernel_size + 1, 2* dilate_kernel_size+1 ), Point(dilate_kernel_size, dilate_kernel_size));
+	
+	//OpenCV does 0-180 instead of 0-360, so just divide by 2
+	red = ColorThresh{Scalar(0, 10, 100), Scalar(30, 255, 255)}; //0 to 30/2 (includes orange)
+	red2 = ColorThresh{Scalar(155, 10, 100), Scalar(180, 255, 255)}; // 330/2 to 180
+	blue = ColorThresh{Scalar(90,100,100),Scalar(150,255,255)}; // 210/2 to 270/2
+	green = ColorThresh{Scalar(30, 85, 25),Scalar(85,255,255)}; // 70/2 to 170/2
```
DSsoto: Lots of magic constants, you should get these from a config file using ros::param.

sensor_control/vision/symbols/src/ShapeDetector.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "ShapeDetector.h"
+
+float ShapeDetector::findAngle(cv::Point p1, cv::Point p2, cv::Point p3) {
+  cv::Point v1 = p2 - p1;
+  cv::Point v2 = p3 - p1;
+  float m1 = sqrt(v1.x * v1.x + v1.y * v1.y);
+  float m2 = sqrt(v2.x * v2.x + v2.y * v2.y);
+  float dot = v1.x * v2.x + v1.y * v2.y;
+
+  float thecos = dot / (m1 * m2);
```
DSsoto: You can just do `v1.dot(v2`) instead of `dot` here.

sensor_control/vision/symbols/src/ShapeDetector.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+float ShapeDetector::findMean(std::vector<float> observed) {
+  float sum = 0;
+  for (int i = 0; i < observed.size(); i++) {
+    sum += observed[i];
+  }
+  return sum / observed.size();
+}
+float ShapeDetector::findVarience(std::vector<float> observed) {
```
DSsoto: variance

sensor_control/vision/symbols/src/ShapeDetector.cpp
```diff
@@ -0,0 +1,167 @@
...
+  return sum / expected;
+}
+
+float ShapeDetector::findMean(std::vector<float> observed) {
+  float sum = 0;
+  for (int i = 0; i < observed.size(); i++) {
+    sum += observed[i];
+  }
+  return sum / observed.size();
+}
```
DSsoto: This would be a lot more succinct:  `float average = std::accumulate( v.begin(), v.end(), 0.0)/v.size();`
accumulate is defined in <numeric>

sensor_control/vision/symbols/src/ShapeDetector.cpp
```diff
@@ -0,0 +1,167 @@
...
+float ShapeDetector::findVarience(std::vector<float> observed) {
+  float mean = findMean(observed);
+  float sum = 0;
+  for (int i = 0; i < observed.size(); i++) {
+    sum += (observed[i] - mean) * (observed[i] - mean);
+  }
+  return sum / (observed.size() - 1);
+}
+
+bool ShapeDetector::boundingAreaCross(std::vector<cv::Point> &points) {
```
DSsoto: Function name is misleading, I would expect it to return the area of something, not a boolean, if I encountered it in source code.

sensor_control/vision/symbols/src/ShapeFind.cpp
```diff
@@ -0,0 +1,73 @@
...
+      symbols->list.push_back(hold);
+    } else if (ShapeDetector::isCircle(shapes[i])) {
+      navigator_msgs::DockShape hold = fillHold(shapes[i], navigator_msgs::DockShape::CIRCLE);
+      symbols->list.push_back(hold);
+    }
+  }
+}
+
+navigator_msgs::DockShape ShapeFind::fillHold(std::vector<cv::Point> shape, std::string symbol) {
+  navigator_msgs::DockShape hold;
```
DSsoto: Why is this called 'hold'?

sensor_control/vision/symbols/src/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+      // Grab ros frame
+      cv_bridge::CvImagePtr cv_ptr;
+      try {
+        cv_ptr = cv_bridge::toCvCopy(msg, sensor_msgs::image_encodings::BGR8);
+      } catch (cv_bridge::Exception &e) {
+        ROS_ERROR("cv_bridge exception: %s", e.what());
+        return;
+      }
+      // Convert Ros frame to opencv
+      cv::waitKey(3);
```
DSsoto: Dont you want to enclose this in `#ifdef DO_DEBUG`?

ironmig:
@mattlangford 
For some reason service/message files are not being built by semaphore. Heres the error:
```
*** Warning, failed to generate libraries for the following packages: ***
    urdf_tutorial (missing dependency)
    kill_handling
    rawgps_common
    navigator_msgs
    skytraq_driver
    sub8_actuator_driver
    nv08c_driver
    sub8_simulation
    magnetic_dynamic_compensation
    uf_common
    rise_6dof
    odom_estimator
    c3_trajectory_generator
    hydrophones
    sub8_msgs
    sub8_gazebo
```

I've run into the problem before when I didn't run
``` sudo rosdep init``` when setting up my workspace and ```rosdep update``` after.

I imagine this is why the shooter isn't building too. Can you add ```rosdep update``` to the semaphore build process and make sure the install script runs ``` sudo rosdep init```?

**Comment by mattlangford** [comment]
The install script runs those two commands when installing things. I will explicity re-execute them and see what happens

File: mission_control/navigator_missions/nav_missions/symbol_mission.py [hunk 13]
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python
+import txros
+import rospy
+import std_srvs.srv
+from navigator_msgs.msg import DockShape
+import numpy
```
mattlangford: It is a standard conversation to `import numpy as np`. This does a couple things, shortens a lot of the code up (since Numpy is used everywhere) and keeps consistency with current design choices.

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+import nav_msgs.msg
+
+import cv2
+
+from twisted.internet import defer
+
+
+
+error_threshold = 0.01
+move_per_iteration = 1.0;
```
mattlangford: Find and Replace `;` ` `

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+#~ square = numpy.array([Point(-1,-1,0), Point(-1,1,0), Point(1,1,0), Point(1,-1,0)])
+#~ square = numpy.array([-1,-1,0,-1,1,0,1,1,0,1,-1,0])
+square = numpy.array([[0, 0, 0],[ 1, 0, 0], [0, 1, 0],[ 1, 1, 0]]).astype(numpy.float64)
+#~ K = numpy.array(numpy.mat("1 0 0;0 1 0;0 0 1"))
+K = numpy.array([[1,0,0],[0,1,0],[0,0,1]]).astype(numpy.float64)
+
+
+def boundingRect(points):
```
mattlangford: According to [PEP8](https://www.python.org/dev/peps/pep-0008/) (the python style guides, and the rules we use for all the python on the boat), function names are supposed to be all lowercase separated by underscores (`bounding_rect(points)`). On the sub we were pretty strict about pep8, and its really helpful for people who have to look at your code to have everything (class, variables, and method names among other things) standardized with the rest of the repo.

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+    maxX = 0
+    minX = 2000
+    maxY = 0
+    minY = 2000
+    for i in range(len(points)):
```
mattlangford: This could be cleaner with:
```
for point in points:
    if point.x > maxX:
...etc...
```
Not a super big issue, but just FYI

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+    return numpy.array([maxX, maxY, minX, minY, minX, maxY, maxX, minY])
+    #~ return [Point(minX, maxY, 0), Point(maxX, minY,0), Point(maxX, maxY,0), Point(minX, minY,0)]
+
+def callback(data):
+    #~ K = numpy.reshape(data.K, (3,3))
+    print K
+    
+@txros.util.cancellableInlineCallbacks
+def listener(navigator):
+    camerainfoSub = yield navigator.nh.subscribe("/side_camera/camera_info", sensor_msgs.msg.CameraInfo)
```
mattlangford: PEP8 calls for no camel case for variables and to separate with underscores (`camera_info_sub`)

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+    caminfo = yield b.get_next_message()
+    #~ K = caminfo.K
+    #print caminfo.K
+    defer.returnValue(caminfo.K)
+    #~ rospy.spin()
+
+@txros.util.cancellableInlineCallbacks
+def main(navigator):
+    
+    #~ shooterFire = rospy.ServiceProxy("/shooter/fire", std_srvs.srv.Trigger)
```
mattlangford: Does your editor do this (`#~`) automatically?

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+    #~ K = yield listener(navigator)
+    #~ yield navigator.nh.sleep(1)
+    position, orientiation = navigator.pose
+    
+    print orientiation
+    
+    print K
+    try:    
+        resp = yield navigator.vision_request("get_shape")
+        error = float(resp.symbol.CenterX)/resp.symbol.img_width - 0.5
```
mattlangford: If you want to avoid having to cast one of these as floats, you can `from __future__ import division` at the top and then it'll automattically do float division rather than integer division.

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+        p2 = boundingRectPoints[1]
+        p2 = numpy.insert(p2, 2, 0)
+    
+        p3 = boundingRectPoints[2]
+        p3 = numpy.insert(p3, 2, 0)
+    
+        p4 = boundingRectPoints[3]
+        p4 = numpy.insert(p4, 2, 0)
+
+        pointA = numpy.dot(p1, rvec1) + numpy.matrix.transpose(tvec)
```
mattlangford: Another FYI, this would be a cleaner line:
`point_a = p1.dot(rvect1) + tvect.T`

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,133 @@
...
+    #~ cameraInfo = rospy.ServiceProxy("/side_camera/camera_info", sensor_msgs.msg.CameraInfo)
+    #~ K = yield listener(navigator)
+    #~ yield navigator.nh.sleep(1)
+    position, orientiation = navigator.pose
+    
+    print orientiation
+    
+    print K
+    try:    
+        resp = yield navigator.vision_request("get_shape")
```
mattlangford: Your service provider script needs to return wether it succeeded or not so we don't have to surround this whole thing in the try statement

mattlangford:
Can we move this to the `dev` branch of the repo?

ironmig:
I think that's a good idea. Not quite "production" ready

DSsoto:
Closing this PR then, just redo it on the dev branch.

mission_control/navigator_missions/nav_missions/symbol_mission.py
mattlangford: Not merging without this, lol

sensor_control/vision/symbols/src/FrameProc.cpp
```diff
@@ -0,0 +1,110 @@
...
+  namedWindow("blue",CV_WINDOW_AUTOSIZE);
+  namedWindow("green",CV_WINDOW_AUTOSIZE);
+  namedWindow("red",CV_WINDOW_AUTOSIZE);
+  createTrackbar("Blur Size","blured",&blur_size,100);
+  #endif
+}
+void FrameProc::init(ros::NodeHandle& nh)
+{
+  //Set HSV values
+  nh.getParam("hsv/red1/low/H",red.low[0]);
```
mattlangford: Consider [retrieving a list](http://wiki.ros.org/roscpp/Overview/Parameter%20Server#Retrieving_Lists) from the ros parameter server instead of individually like this - it would save a lot of space.

mission_control/navigator_missions/nav_missions/3Ddetect.py
```diff
@@ -0,0 +1,210 @@
...
+        if(points[i][1] > maxY):
+            maxY = points[i][1];
+        if(points[i][0] < minX):
+            minX = points[i][0];
+        if(points[i][1] < minY):
+            minY = points[i][1];
+    return np.array([[maxX, maxY], [minX, minY],[maxX,minY],[minX,maxY]])
+
+@txros.util.cancellableInlineCallbacks
+def main(navigator):
```
mattlangford: This looks like it's entirely test code - it shouldn't be in the master repo IMO

mission_control/navigator_missions/nav_missions/symbol_mission.py
```diff
@@ -0,0 +1,96 @@
...
+  print "Loading"
+  shooterLoad()
+
+  resp = yield navigator.vision_request("get_shape")
+  center = float(resp.symbol.CenterX) / resp.symbol.img_width
+  error = center - 0.5
+  while abs(error) > error_threshold:
+    if error < 0:
+        print "Turning Left"
+        yield navigator.move.yaw_left(0.1).go()
```
mattlangford: You can yaw_left by a negative amount 
`yeild navigator.move.yaw_right(np.sign(error) * .01).go()` for example

sensor_control/vision/symbols/src/ShoreDebug.cpp
```diff
@@ -0,0 +1,162 @@
...
+    it_(nh)
+  {
+    red = ColorThresh{Scalar(0,10, 100), Scalar(30, 255, 255)};
+    red2 = ColorThresh{Scalar(155, 10, 100), Scalar(180, 255, 255)};
+    blue = ColorThresh{Scalar(90,100,100),Scalar(150,255,255)};
+    green = ColorThresh{Scalar(30, 85, 25),Scalar(85,255,255)};
+    initParams();
+    image_sub_ = it_.subscribe(camera_topic, 1, &ShoreDebug::newFrameCallback, this);
+    foundShapesSubscriber = nh.subscribe("/dock_shapes/found_shapes", 1000, &ShoreDebug::foundShapesCallback, this);
+    namedWindow("Result",CV_WINDOW_AUTOSIZE);
```
mattlangford: What we did on the sub is have the perception nodes publish debug images as they were working. So instead of having a specific debug class, the main node would be publishing out debug images and information so that you could view it in rqt

sensor_control/vision/symbols/src/FrameProc.cpp
ironmig: If I had to redo it, I would probably do it that way. But might as well leave it now that's it's already done. Don't have to remember which order it is in

sensor_control/vision/symbols/src/ShoreDebug.cpp
ironmig: I kinda like having a separate debug program so I can add whatever I want to it without effecting the real node's performance/behavior.

mission_control/navigator_missions/nav_missions/symbol_mission.py
ironmig: We really have to think of a different way to do this altogether. What do you think about some sort of basic controller that takes the error and turns a certain speed based on it? PID or something? I feel like the jerk left 10 degrees in a loop is bad

mattlangford: By not yielding on the spinnning, you can rotate the boat while looking for the shape at the same time and stop when you see it - this is a benefit of using txros. The safest way to do this is to set a waypoint 180 degrees from the current heading, and then set another one once you get there. Having a third source of control on the boat could get finicky and seems a bit redundant.

sensor_control/vision/symbols/src/ShoreDebug.cpp
mattlangford: But couldn't your debug code and your actual code diverge such that your debug screen doesn't show what the boat is acting on? The performance hit of publishing an extra couple of images is very minimal and it doesn't use the network unless you're looking at the feed.

DSsoto: Also, the way we dd it on the sub, which proved to be very effective, was encasing the code for creating the debug images and publishing them inside conditionally compiled blocks like so:
 ```
#ifdef VISUALIZE
  visualization_code_goes_here...
#endif
```

**Comment by mattlangford** [comment]
Any updates here?

**Comment by ironmig** [comment]
At this point we're waiting for lidar so we can align to the target once found. @RustyBamboo plans on changing the mission to circle around the 4-sided target. Is there an easy way to do this (navigator.move.circle() or something)?

The vision itself might need some additional tweaking. I'll look at bags today to see where it fails (if I can find out how to access them...)

**Comment by mattlangford** [comment]
Check my newest PR, it's got a circle point method in the pose editor class, if you need help with implementation, let me know.

**Comment by mattlangford** [comment]
Make sure @rleyva knows that you're wanting the to have the lidar move us to the tower

File: mission_control/navigator_missions/nav_missions/detect_deliver.py [hunk 26]
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python
+import txros
```
mattlangford: There are a couple unused imports here

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+    return np.array([maxX, maxY, minX, minY])
+
+
+def area_of_rect(rect):
+    return (rect[1] - rect[2]) * (rect[1] - rect[3])
+
+
+@txros.util.cancellableInlineCallbacks
+def main(navigator):
+  shooterLoad = rospy.ServiceProxy("/shooter/load", std_srvs.srv.Trigger)
```
mattlangford: Does this work without `rospy.init()`?

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+  shooterFire = rospy.ServiceProxy("/shooter/fire", std_srvs.srv.Trigger)
+  
+  resp = yield navigator.vision_request("get_shape")
+  if not resp.found and resp.symbol.img_width != 0:
+    navigator.move.yaw_left(360, "deg").go();
```
mattlangford: No semicolon, and is this going to do anything? If we set a waypoint at 360 degrees, it'll be in the same position as we are now.

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+    while not (resp.found and resp.symbol.img_width != 0):
+      resp = yield navigator.vision_request("get_shape")
+    yield navigator.move.go() #Stop moving once shape seen
+
+  print "Found"
+  print "Loading"
+  shooterLoad()
+
+  resp = yield navigator.vision_request("get_shape")
+  center = float(resp.symbol.CenterX) / resp.symbol.img_width
```
mattlangford: You don't have to cast these to float every time if you add `from __future__ import division`

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+  error = center - 0.5
+  if abs(error) > error_threshold:
```
mattlangford: This seems redundant, haven't we already found it?

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+      resp = yield navigator.vision_request("get_shape")
+      center = float(resp.symbol.CenterX) / resp.symbol.img_width
+      error = center - 0.5
+    yield navigator.move.go() #Stop moving once centered
+  print "Center Proportion: ", center
+  print "Centerted"
+  
+  resp = yield navigator.vision_request("get_shape")
+  rect = bounding_rect(resp.symbol.points)
+  width = (rect[0] - rect[2]) / resp.symbol.img_width
```
mattlangford: Make sure you do that future import, or this might be calculated using integer division

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -0,0 +1,105 @@
...
+    yield navigator.move.go() #Stop moving once centered
+
+  print "Correct Distance"
+  yield navigator.move.go()
+  
+  print "Shooting"
+  shooterFire()
+
+  #shoots another 3 times
+  for i in range(0,3):
```
mattlangford: You don't need to include the 0, so the line would become `for i in range(3):`

mission_control/navigator_missions/nav_missions/3Ddetect.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python
```
mattlangford: I think this approach is a better one compared to trying to center the object in the frame since that does not account for rotations or anything like that. The best solution would be to use the lidar to produce a normal vector for the surface and line up with that. But having the 3d pose of the shape is still be preferred
That being said, this file should remain in a feature branch and not here in master

mattlangford:
Please address comments, this needs to be in the master branch

mattlangford:
There are still some issues in the mission - but you can get those fixed later

rleyva:
Hey Kevin, could you fix the build issues ASAP? We need to merge it in for this Saturday.

ironmig:
Are there still build issues? I wasn't aware...

You can certainly merge this, but be aware that the vision is not very reliable. In ideal conditions it operates okay, but @RustyBamboo and I are considering different options for a different solution.

mattlangford:
Yeah both Ralph and I tried building on our computers and it failed on both during the message generation still. We tried the fixes listed on the issue but no luck

ironmig:
I'm so confused... are the build errors on #41, #60 or both?

mission_control/navigator_missions/nav_missions/detect_deliver.py
ironmig: Yes in testing it has. I  guess the navigator singleton does that?

mission_control/navigator_missions/nav_missions/detect_deliver.py
ironmig: commented out for now

mission_control/navigator_missions/nav_missions/detect_deliver.py
ironmig: It's possible that we lose it while moving

ironmig:
Notes recent changes:

- Created an abstract class for anything that can be given a frame  and find shapes so we can test different methods
- Added a method similar to the old one, but using grayscale instead of thresholding colors
- After finding a shape, it looks at where it is in the color image to determine the color
- Uses a cropped portion of the frame so it doesn't see the protective case and the clouds
- Detection of shapes in grayscale method not very good, works well with circles, poorly with triangles/cross
- Need to do more geometry checks to filter out false positives/find correct shapes
- Change DO_DEBUG compiler variable in Grayscalecontour.h to remove the debug windows/sliders
- Detect Deliver mission assumes shape already found, rotates to center, moves towards/away to get correct distance, then shoots 4 balls
- Lots of magic numbers in DetectDeliver mission and Grayscalecontour vision method that may need to be tuned
- PLEASE get lots of bags with all shapes/colors with camera calibrated and whitebalance/other camera settings locked down. Get bags from very close, somewhat close, and far away
- Get bags where multiple shapes can be seen at once
- I'm goinna sleep

mission_control/navigator_missions/nav_missions/detect_deliver.py
forrestv: Your really shouldn't be mixing rospy and txros code. The equivalent in txros is:

    shooterLoad = nh.get_service_client("/shooter/load", std_srvs.srv.Trigger)
    ...
    yield shooterLoad(std_srvs.srv.TriggerRequest())

If this actually works, it means something else is calling `rospy.init()`, which is its own problem. If you use both rospy and txros, your node will appear as two different nodes, which is probably undesirable.

ironmig:
Can be merged (git wise) now. I think all the concerns were addressed or are minor enough to still have in dev

sensor_control/vision/symbols/src/ShoreDebug.cpp
ironmig: added in  	60ad6d2

ironmig:
MEerrrrggeeee

mission_control/navigator_missions/nav_missions/detect_deliver.py
ironmig: @forrestv The shooter is now going to use actionlib for its control. How can I create an actionclient in txros?

mattlangford: Look at pose editor - it has the movement action client which is an action.

DSsoto:
Good job @ironmig and @RustyBamboo!

PR #44, opened 2016-09-03
RC Fixes, Start Gate Mission, Pre-Flight Check

mattlangford:
It includes lots of other people's stuff since this is a behind branch. In the future, this won't be a problem - not sure how git will handle that when merging into master.


gnc/navigator_perception/nodes/start_gate_manual.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def mouse_cb(self, event, x, y, flags, param):
+        self.draw_img = np.copy(self.img)
+        self.mouse_position = (x, y)
+
+        if event == cv2.EVENT_LBUTTONUP:
+            self.point = np.array([x, y])
+
+
+def intersect(A, a, B, b):
```
mattlangford: I plan on redoing this using the opencv stereo coorisponder

DSsoto:
Matt pull requests should never have 61 modified files!!

DSsoto:
I guess from now on we'll have light code reviews for merging to dev and then more in depth ones after the code has been tested at pool days and take into account code cleanliness and readability as well as correctness before merging to master.

PR #47, opened 2016-09-08
SHOOTER: Update controller for new feeding mechanism

ironmig:
This PR contains changes for the firmware to be run on the arduino to control the shooter. Because the shooter lacks any sensor feedback, the controller is based on timing (tuned in lab), which can be configured at runtime with the yaml file.


mission_systems/shooter/launch/config.yaml
```diff
@@ -0,0 +1,8 @@
+controller:
```
mattlangford: Good, this is the right way to do this.

mission_systems/shooter/firmware/shooter.cpp
```diff
@@ -223,46 +351,56 @@ class Comms
...
+      res.success = true;
+    }    
+    #endif
+
   public:
     Comms() :
-      str_msg(),
-      //~ sub("/shooter/control",&messageCallback),
+      #ifdef USE_LINEAR_FEEDER
+      fireService("/shooter/fire", &fireCallback),
```
mattlangford: Can you put comments in here as to what does what? One time, Daniel and I spent 10 or 15 minutes trying to figure out how to use all the service calls when we were testing.

mission_systems/shooter/firmware/shooter.cpp
```diff
@@ -150,6 +151,122 @@ class Pololu : public SpeedController
...
+          runLoad();
+          break;
+        case 2:
+          runFire();
+          break;
+      }
+    }
+
+};
+unsigned long AutoController::SPIN_UP_TIME = 1000; //Time to spin up flywheels before feeding balls in
```
mattlangford: Is this not in the yaml config?

mission_systems/shooter/firmware/shooter.cpp
```diff
@@ -150,6 +151,122 @@ class Pololu : public SpeedController
...
+      if (cur_time < RETRACT_TIME)
+      {
+        feeder.reverse();
+      } else if (cur_time < (RETRACT_TIME + MID_LOAD_PAUSE ) )
+      {
+        feeder.off();
+      } else if (cur_time < (RETRACT_TIME + MID_LOAD_PAUSE + LOAD_TIME) ) {
+        feeder.on();
+      }else {
+				feeder.off();
```
mattlangford: Weird spacing here?

ironmig:
Added some documentation to both the README and the top of the firmware file. 

The timing values are in the yaml, but I went ahead and put in some defaults.

ironmig:
After a firm talking to by Mr. Dugger, I've decided to just comment out the part of the CMAKE that compiles the arduino firmware whenever catkin runs. This was causing build issues for pretty much everyone... at it really doesn't have to be built ever once the arduino is programmed.

PR #53, opened 2016-09-12
Remove outdated, unused navigator_state_estimation package

forrestv:


mattlangford:
For those wondering, this package is not used by the boat. All the state estimation code is ran from elsewhere

PR #56, opened 2016-09-14
Lake Day 9-11

mattlangford:


mattlangford:
Good to merge this?

DSsoto:
This introduces changes that are overwritten by other pull requests that are still open, since it would change the base for some of those later changes, wouln't it it make all the other PR's unmergeable, at least automatically. 

Maybe we should merge this after the other big PR's we have open now and use the merge option -Xtheirs

mattlangford:
What specifically? If we merge this then you rebase from the new master things should be fine no?

PR #59, opened 2016-09-17
Mission and Sim work

mattlangford:
- I added a Searcher class. The idea here is that you pass in an arbitrary search pattern defined with pose editor movements, then supply a vision service request type and it'll preform the search until the service returns back (the service needs to have a `found` attribute)
- @DSsoto had merged all of the alarm code from the sub to here. This may not be the complete set of changes, but he will merge those in once they are finished.
- I did some work with color thresholding to find the buoys for a demo for Andy, I only plan to use this as a back up on the next lake day if I can't get the lidar based start gate mission done.
- @ironmig @RustyBamboo the circle code requires a 3d point in space to rotate around - I'm not sure if you were generating that or were depending on the lidar to be your north star
- The TF launch here should be closer to accurate ~~maybe? If not I will add it tomorrow before testing.~~
- Forrest's addition here is a clock publisher and launch file that are launched on the boat startup. This will make sure all ROS components are synced up (TF). __IMPORTANT:__ Make sure you get the time from ROS anytime you use time - don't use the system time as the two could be different.
- There are a bunch of mission changes made on 9-18, so if you're wanting to learn about the mission, I would take a look at those because there are some cool tricks I do in there.
- Also brings Anthony's changes into `dev`


mission_control/navigator_missions/navigator_singleton/pose_editor.py
```diff
@@ -143,58 +144,65 @@ def yaw_left(self, angle, unit='rad'):
...
+        ecef_pos, enu_pos = self.nav.ecef_pose[0], self.nav.pose[0]
 
-    def zero_roll_and_pitch(self):
-        return self.set_orientation(look_at_without_pitching(self.forward_vector))
+        # These functions want radians
+        lat, lon = np.radians([lat, lon], dtype=np.float64)
+        ecef_vector = ecef_from_latlongheight(lat, lon, alt) - ecef_pos
+        enu_vector = enu_from_ecef(ecef_vector, ecef_pos)
+        enu_vector[2] = 0  # We don't want to move in the z at all
+        return self.set_position(enu_pos + enu_vector)
```
mattlangford: @forrestv I think this should work, do you agree?

forrestv: Looks good to me.

RustyBamboo:
For Sunday we could make the circle code with GPS points?

mattlangford:
If the GPS waypoints work out well then it should be possible

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -22,33 +23,40 @@ def __init__(self, nh):
...
 
         defer.returnValue(self)
 
     @property
-    def pose(self):
-        last_odom_msg = self._odom_sub.get_last_message()
-        return odometry_to_numpy(last_odom_msg)[0]
+    @util.cancellableInlineCallbacks
+    def tx_pose(self):
+        last_odom_msg = self._odom_sub.get_next_message()
```
mattlangford: Need a yield here

DSsoto: @mattlangford need a yield here

mattlangford:
This is up to date with latest merge and fixed @DSsoto's astute observation

mission_control/navigator_launch/launch/subsystems/cameras/calibration/00b09d0100e84a42.yaml
```diff
@@ -0,0 +1,20 @@
...
+  cols: 5
+  data: [-0.1701, 0.1220, -.00085339, -0.0012, -0.0304]
+rectification_matrix:
+  rows: 3
+  cols: 3
+  data: [0.9996475214662156, 0.003230347798860112, 0.02635142651832315, -0.003259850357149076, 0.9999941070295705, 0.001076699484903968, -0.02634779311633482, -0.001162221678595386, 0.9996521610233574]
+projection_matrix:
+  rows: 3
+  cols: 4
+  data: [702.1922, 0, 476.5935, -40.00, 0, 699.2620, 298.8085, 0, 0, 0, 1, 0]
```
DSsoto: @rleyva, the opencv calibration that I did about a month ago, which had the camera intrinsics corroborated by the matlab calibration tool and are really close to the intrinsics you have here, and which also correctly backed out the distance between the cameras and almost identical orientations,  say that the fourth coefficient here should be -140, not -40. 
Look at the [calibration file](https://github.com/uf-mil/Navigator/blob/dev/gnc/navigator_perception/calibration.xml). It has been on Navigator for a while now (although not on the repo) and Tess and I have been using it successfully for stereo stuff.

mattlangford: Just fixed, thanks for the info

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -74,6 +85,101 @@ def _load_vision_serivces(self, fname="vision_services.yaml"):
...
+        else:
+            print "SEARCHER - There was an error."
+            print failure.printTraceback()
+            # Handle error
+
+    @util.cancellableInlineCallbacks
+    def start_search(self, timeout=60, loop=True, spotings_req=2, speed=.1):
+        print "SERACHER - Starting."
+        looker = self._run_look(spotings_req).addErrback(self.catch_error)
+        searcher = self._run_search_pattern(loop, speed).addErrback(self.catch_error)
```
DSsoto: Looker and searcher sound really similar, to make this a little easier to read, I suggest maybe finder instead of looker and mover instead of searcher.

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -58,19 +59,12 @@ def tx_ecef_pose(self):
     def move(self):
         return PoseEditor2(self, self.pose)
 
-    def change_wrench(self, source):
-        return self._change_wrench(navigator_srvs.WrenchSelectRequest(source))
-
     def vision_request(self, request_name, **kwargs):
-        assert request_name in self._vision_proxies.keys(), "Unknown request: {}".format(request_name)
+        print "DEPREICATED: Please use new dictionary based system."
```
DSsoto: DEPRECATED*

PR #65, opened 2016-09-22
Scan the code simulation fix

tessbianchi:
Fixes merging issue from #58

### DIFF
--- a/simulation/navigator_gazebo/src/animated_box.cc
+++ b/simulation/navigator_gazebo/src/animated_box.cc
@@ -1,59 +1,53 @@
 //#include <QDebug>
 
 #include "animated_box.h"
-using namespace gazebo;
 
 // Register this plugin with the simulator
 GZ_REGISTER_VISUAL_PLUGIN(ModelPush)
 
-void ModelPush::Load(rendering::VisualPtr _parent, sdf::ElementPtr _sdf)
-{
-    std::cout << "loading plugin, animated_box..." << std::endl;
-    // Store the pointer to the model
-    GZ_ASSERT(_parent != NULL, "Received NULL model pointer");
+void ModelPush::Load(gazebo::rendering::VisualPtr _parent,
+                     sdf::ElementPtr _sdf) {
+  // Store the pointer to the model
+  GZ_ASSERT(_parent != nullptr, "Received NULL model pointer");
 
-    this->model = _parent;
-    // Listen to the update event. This event is broadcast pre-render update???
-    this->updateConnection = event::Events::ConnectPreRender(boost::bind(&ModelPush::OnUpdate, this));
-    begin_time = clock();
+  this->model = _parent;
+  // Listen to the update event. This event is broadcast pre-render update???
+  this->updateConnection = gazebo::event::Events::ConnectPreRender(
+      boost::bind(&ModelPush::OnUpdate, this));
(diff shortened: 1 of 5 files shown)


simulation/navigator_gazebo/src/animated_box.cc
```diff
@@ -0,0 +1,53 @@
...
+      c = gazebo::common::Color(0.0, 0.0, 0.0);
+    }
+    begin_time = clock();
+    count = (count + 1) % 4;
+
+  }
+
+  else if (secs > 1.9) {
+    c = gazebo::common::Color(1.0, 0.0, 0.0);
+    begin_time = clock();
```
DSsoto: This wouldn't work right because you're using system time. If you do `ros::Time begin_time = ros::Time::now();`, it will handle sim time correctly.

DSsoto: @tessbianchi Sorry, Wrong PR, but still applies if its the same on your new one.

simulation/navigator_gazebo/src/animated_box.h
```diff
@@ -0,0 +1,26 @@
+#ifndef ANIMATED_BOX_H
+#define ANIMATED_BOX_H
+
+#include <boost/bind.hpp>
+#include "gazebo/gazebo.hh"
+#include "gazebo/rendering/Visual.hh"
+#include "gazebo/common/common.hh"
+#include <ctime>
```
DSsoto: Don't forget to remove the include.

simulation/navigator_gazebo/src/animated_box.cc
```diff
@@ -0,0 +1,53 @@
...
+
+  if (count != 0 && secs > .9) {
+    if (count == 1) {
+      c = gazebo::common::Color(0.0, 1.0, 0.0);
+    }
+    if (count == 2) {
+      c = gazebo::common::Color(0.0, 0.0, 1.0);
+    }
+    if (count == 3) {
+      c = gazebo::common::Color(0.0, 0.0, 0.0);
```
tessbianchi: It doesn't really matter because it just needs to be a repeating sequence of colors, it doesn't matter if its red, green, blue or blue green red

simulation/navigator_gazebo/src/animated_box.cc
tessbianchi: This works well.  but I will change it.

tessbianchi: This shouldn't matter because:

float secs = float(clock() - begin_time) / CLOCKS_PER_SEC;

tessbianchi: I'm dividing by clocks per second, which should give me something that is close enough to what I want.

tessbianchi:
Should be good to go!

simulation/navigator_gazebo/src/animated_box.cc
tessbianchi: Actually I'ma change it

tessbianchi:
Ok, now it should be good!

whispercoros:
@tessbianchi, Do you mind if I change the element so that it has the three faces specified by the competition document?

tessbianchi:
That would be great!

PR #66, opened 2016-09-22
Lake Day 9-18

mattlangford:
Note there are a lot of commits in here that broke the build - those were fixed in a commit here


DSsoto:
These have been tested at a lake day, and @mattlangford and I have just tested that it builds, merging to master.

PR #67, opened 2016-09-23
Partial redo of reverted redo of PR57

DSsoto:


DSsoto:
Can confirm that every single one of these commits were made by me except for three that @mattlangford made to a branch of my fork. This should not overwrite anyone's work and these commits have been previously reviewed.

PR #69, opened 2016-09-26
Updates from lake day

### LINKED ISSUE #45: Thrusters not giving feedback
Not sure if it's a hardware, firmware, or software issue. The thrusters won't return feedback information (we mainly just want battery voltage)

mattlangford:
This should fix the "17 commits behind" thing in dev.

Also fixes #45 - albeit it's a patch at best.

PR 69 is best PR


mattlangford:
@whispercoros make sure the install script this would put into dev is correct

whispercoros:
This is not correct at all, both the readme file and install script were reverted to old versions. There are even two copies of the readme in this branch on your fork.

mattlangford:
Well shoot, I'll try to fix that tomorrow

DSsoto:
In the future try to limit the amount of changes per pull request.

PR #70, opened 2016-09-30
Lake Day 9-24

mattlangford:
Reintroduces stereo changes into master along with lake changes from 9-24
@whispercoros make sure this doesn't mess up the install script


whispercoros:
None of these changes mess with the script or readme and it builds, so this looks good to merge

PR #71, opened 2016-09-30
SHOOTER: Work with new motor controller + other improvements

ironmig:
* Removed all arduino stuff (new motor controller operates directly over USB to main PC)
* Shooter load/fire are now actionlib scripts
 * Will abort if invalid (double loading, firing without loading)
 * Receive feedback about when load/fire will be completed
 * Cannot try to do two things at once
* Python class to interface with Sabertooth 2x12 (motor.set(1), etc) using pyserial


mission_systems/shooter/nodes/Sabertooth2x12.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
mattlangford: I'm glad you're using python for this - however I would take a look at other python scripts (or look at the [PEP8](https://www.python.org/dev/peps/pep-0008/) guidlines) in the navigator or sub repo to get an idea of how things should be formatted. Alternatively you could get a linter that will look at your code as your write it and show you were the PEP errors are.

A couple things right off the bat, I think your python scripts could use more whitespace. PEP8 calls for a new line after every class function. Next, __init__() generally goes at the top of a class so anyone looking at your code knows exactly what the class takes as parameters and knows where to start.

mission_systems/shooter/nodes/Sabertooth2x12.py
```diff
@@ -0,0 +1,42 @@
...
+    send_char = self._map(contrained_num,-1.0,1.0,128,255)
+    byts = struct.pack("B",send_char)
+    self.ser.write(byts)
+    self.ser.flush()
+  def getMotor2(self):
+    return self.motor_2_speed
+  def __init__(self,filename):
+    self.motor_1_speed = 0.0
+    self.motor_2_speed = 0.0
+    self.filename = filename
```
mattlangford: You're not using this anywhere else so it probably doesn't need to be a class variable - just pass it directly into the serial port

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -0,0 +1,128 @@
...
+    if (self.load_server.is_active()):
+      self.load_server.set_aborted(result)
+    if (self.fire_server.is_active()):
+      self.fire_server.set_aborted(result)
+    time.sleep(0.1)
+    self.motor_controller.setMotor1(0)
+    self.motor_controller.setMotor2(0)
+    rospy.loginfo("canceled")
+  def __init__(self):
+    self.controller_file = rospy.get_param('~controller/controller_serial', '/dev/ttyUSB0')
```
mattlangford: Make sure this param gets set somewhere in the launch (it looks like it got taken out), else you should probably set the serial port to use the `by-id` path by default.

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -0,0 +1,128 @@
...
+    rospy.loginfo("Finished fire")
+  def cancelCallback(self,req):
+    result = ShooterDoActionResult()
+    result.result.success = False
+    result.result.error = result.result.MANUAL_CONTROL_USED
+    if (self.load_server.is_active()):
+      self.load_server.set_aborted(result)
+    if (self.fire_server.is_active()):
+      self.fire_server.set_aborted(result)
+    time.sleep(0.1)
```
mattlangford: Try not to mix ros time with normal python time - especially on the boat where some cooky things go on

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
```
mattlangford: Same formatting comments as above

mission_systems/shooter/nodes/ShooterControl.py
ironmig: It is in the yaml

ironmig:
MERGGE

mission_systems/shooter/nodes/ShooterControl.py
ironmig: Fixed in  89f07f3

mission_systems/shooter/nodes/Sabertooth2x12.py
ironmig: 99c475f

mission_systems/shooter/nodes/ShooterControl.py
ironmig: Still need to set by-id once I see what it is on the boat

ironmig:
Ready to merge except for:

- Set serial path to a by-id (can't determine until run on boat, I think)
- Tune timing (should be okay)
- Add manual control service (not really needed)

mattlangford:
Not approving until you appreacite The Imitation Game

mgriessler:
Assuming the mechanical system allows a full retract don't try to ride the margin on timing for a full retract. Build in a healthy safety factor.

PR #73, opened 2016-10-01
GPS Bounds and Coordinate Conversion

mattlangford:
Added services to:
 - Convert between different global frames (from `enu`, `lla`, or `ecef` to all the others. 
 - Request the current GPS bounds in whatever global frame you want

Note, both the coordinate conversion script and bounds server script must be running to have access to either of these services

### DIFF
--- a/mission_control/navigator_missions/navigator_singleton/__init__.py
+++ b/mission_control/navigator_missions/navigator_singleton/__init__.py
@@ -2,7 +2,7 @@
 import os
 
 for module in os.listdir(os.path.dirname(__file__)):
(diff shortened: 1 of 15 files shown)


mission_control/navigator_missions/navigator_singleton/__init__.py
```diff
@@ -2,9 +2,9 @@
 import os
 
 for module in os.listdir(os.path.dirname(__file__)):
-    if module == '__init__.py' or module[-3:] != '.py':
+    if module == '__init__.py' or module[-3:] != '.py' and module[-1] != '~':
```
tessbianchi: GEDIT!

PR #77, opened 2016-10-08
SHOOTER: Tuned timing vals/bug fixes

ironmig:
Longer tract + extend time to avoid jams

cancel and manual callback dont throw error

Serial port set to by-id (not sure if this will be same on boat)


mattlangford:
Looks good

PR #78, opened 2016-10-08
T-GEN: Add ros interface

mattlangford:
This is the first version of the ROS interface for Jason's [lqRRT](https://github.com/jnez71/lqRRT) to NaviGator. There are some planned updates to improve "getting trapped" and other behaviors that will get committed as they get devolved. 

Some cool shots of it working with a simulated occupancy grid:

**Randomly driving between points**
path1

**Randomly driving between points**
path2

**Added obstacles to block trajectory**
path3

There's some code in there that lets me simulate the interface without having to tax my computer with gazebo (namely setting the state to the trajectory).

Also to run this you have to have the lqRRT package installed.


gnc/navigator_trajectory_generation/nodes/jdog.py
```diff
@@ -0,0 +1,409 @@
...
+        for ID in xrange(self.planner.tree.size):
+            traj = states[ID]
+            q = tf.transformations.quaternion_from_euler(0, 0, traj[2])
+            pose_array_list.append(navigator_tools.numpy_quat_pair_to_pose(np.array([traj[0], traj[1], 0]), q))
+
+        pose_array.poses = pose_array_list
+        self.tree_vis_pub.publish(pose_array)
+
+    def path_still_feasible(self):
+        for i, (x, u) in enumerate(zip(self.planner.x_seq, self.planner.u_seq)):
```
tessbianchi: What is the x and u ?

gnc/navigator_trajectory_generation/nodes/jdog.py
```diff
@@ -0,0 +1,409 @@
...
+        states = np.copy(self.planner.tree.state)
+        for ID in xrange(self.planner.tree.size):
+            traj = states[ID]
+            q = tf.transformations.quaternion_from_euler(0, 0, traj[2])
+            pose_array_list.append(navigator_tools.numpy_quat_pair_to_pose(np.array([traj[0], traj[1], 0]), q))
+
+        pose_array.poses = pose_array_list
+        self.tree_vis_pub.publish(pose_array)
+
+    def path_still_feasible(self):
```
tessbianchi: I've been looking at this method, and does it send a replan method to the thread queue if the boat is on track for a collision?

If that is the case, you should rename it to something like ' avoid_impact_replan' or something like that. 'plan_still_feasible' isn't super descriptive

gnc/navigator_trajectory_generation/nodes/thread_queue.py
```diff
@@ -0,0 +1,74 @@
...
+    r = lambda: funct(50, "a")
+    r1 = lambda: funct(30, "b")
+
+    tq = ThreadQueue()
+
+    def add_thread(msg, tq, f):
+        print "Adding thread: ", f, rospy.Time.now() + rospy.Duration(msg.data)
+        tq.put(f, rospy.Time.now().to_sec() + msg.data)
+
+    rospy.init_node("bleb")
```
tessbianchi: bleb

gnc/navigator_trajectory_generation/tools/ogrid_gen.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python
+import rospy
```
tessbianchi: This is super cool!!

In the future, you should add a command line -h option people easily know how to use it.

mattlangford:
Canceling due to updates in lqrrt.

PR #79, opened 2016-10-10
T-GEN: Add ros interface

mattlangford:
This includes Jason's [lqRRT](https://github.com/jnez71/lqRRT) ros node with some additions to make it work with Navigator (using an action client mainly). 

This also removes some old deprecated packages and adds a cool tool to go to rviz waypoints

__TODO__ We need to add the lqrrt repo as a dependency to the Navigator repo. (@whispercoros)


gnc/navigator_trajectory_generation/nodes/behaviors/direct.py
```diff
@@ -0,0 +1,92 @@
+"""
```
DSsoto: Convert tabs to spaces for this whole file and others like it

gnc/navigator_trajectory_generation/nodes/behaviors/params.py
```diff
@@ -0,0 +1,110 @@
...
+							   [ 1.6000,  0.6000, -0.0123],
+							   [ 1.6000, -0.6000, -0.0123]])
+
+# Thruster heading vectors, [back-left, back-right, front-left front-right] (m)
+thruster_directions = np.array([[ 0.7071,  0.7071,  0.0000],
+								[ 0.7071, -0.7071,  0.0000],
+								[ 0.7071, -0.7071,  0.0000],
+								[ 0.7071,  0.7071,  0.0000]])
+
+# Fresh sprinkles
```
DSsoto: +1

gnc/navigator_trajectory_generation/nodes/lqrrt_node.py
```diff
@@ -0,0 +1,554 @@
...
+        self.free_radius = npl.norm(behaviors.fcar.goal_buffer[:2])
+        self.stuck = False
+        self.collision_warning = False
+
+        # Timing stuff
+        self.busy = False
+        self.reevaluating = False
+        self.time_remaining = 0
+        self.get_rostime = lambda: rospy.Time.now().to_sec()
+        self.last_update_time = self.get_rostime()
```
DSsoto: Should this be initialized to `None`?

gnc/navigator_trajectory_generation/tools/client_goal_gen.py
```diff
@@ -0,0 +1,40 @@
...
+
+import txros
+from txros import action
+from twisted.internet import defer
+
+import numpy as np
+import navigator_tools
+from navigator_msgs.msg import MoveToWaypointAction, MoveToWaypointActionGoal, MoveToWaypointActionResult
+
+print_t = navigator_tools.print_t
```
DSsoto: `from navigator_tools import print_t` instead of line 9 and you don't need this

mission_control/navigator_motion_planner/nodes/explorer.py
```diff
@@ -0,0 +1,137 @@
...
+        self.has_control = False
+        self.move_result = None
+
+        rospy.Subscriber("/odom", Odometry, self.got_odom)
+        rospy.Service("/explorer/explore", SetBool, self.search_request)
+
+        #while self._nav_dummy.pose is None and not rospy.is_shutdown():
+        #    rospy.sleep(.1)
+
+        print "Taking control"
```
DSsoto: Doesn't say what's taking control. If it was only for debugging it should be removed.

utils/navigator_msgs/action/MoveToWaypoint.action
```diff
@@ -0,0 +1,13 @@
+# Define a gaol message
```
DSsoto: goal

utils/navigator_tools/navigator_tools/__init__.py
```diff
@@ -4,10 +4,12 @@
...
 import rviz_helpers
+import general_helpers
 
 from init_helpers import *
 from image_helpers import *
 from msg_helpers import *
 from threading_helpers import *
 from geometry_helpers import *
 from rviz_helpers import *
+from general_helpers import *
```
DSsoto: It feels redundant to have both this and line 7. What's the reason for this?

DSsoto:
Really minor stuff, this is all really good.

RIP C3

Also, I would'be been a hypocrite to make comments about you doing regular printing instead of ros logging because I do the same still, but I still think we should move in that direction since its easy to set different priorities to different messages and select what you want to see.

PR #80, opened 2016-10-11
SHOOTER: Fatal syntax bug fix

ironmig:
This is what caused the shooter to not shoot on 10/09


whispercoros:
The changes look very straightforward, so I will merge this in.

ironmig:
Good question... I am confused why too. In lab, I was manually publishing to the goal topic, which I guess didn't cause it to publish the feedback/result where the bug is. The bug happened when the action service was properly called by the mission through txros (which does all the feedback/success stuff). idk

PR #83, opened 2016-10-12
SIM: Sandbox Simulator Part 1

whispercoros:
I am splitting up the simulator sandbox into two separate PRs due to popular demand. This PR contains functional versions of the following challenges:
- Demonstrate Navigation and Control
- Find Totems and Avoid Obstacles
- Identify Symbols and Dock
- Scan the Code
- Acoustic Pinger-Based Transit (actual pinger not yet implemented)

In the next PR, I will be making changes to complete the following challenges:
- Scan the Code (adding the platform and three faces)
- Identify Symbols and Dock (fixing the lighting issues)
- Detect and Deliver
- Coral Survey
- Find the Break

This list is currently in an order governed by increasing difficulty to implement. I would prefer that it be in an order governed by priority, so feedback as to what needs to be tested sooner rather than later would be nice.


simulation/navigator_gazebo/launch/goose.launch
```diff
@@ -2,47 +2,425 @@
...
       -x 0 -y 0 -z 1.5 -model wamv"
     output="screen" />
 
-  <!-- Shapes -->
-  <node name="spawn_blue_cross" pkg="gazebo_ros" type="spawn_model"
-    args="-file $(find navigator_gazebo)/models/shapes/blue_cross.sdf -sdf
-      -x 4 -y -18 -z 1.5 -model blue_cross"
+
+  <!-- Demonstrate Navigation and Control -->
+  <node name="spawn_speed_start_green" pkg="gazebo_ros" type="spawn_model"
```
mattlangford: Can you put a flag in here so that gazebo can be launched like: `roslaunch navigator_launch simulation minimal:=true` and it won't spawn all the obstacles?

mattlangford: Also, you might look into just moving this into a python script that can randomly spawn things (that way this file isn't 5 miles long) (The first comment is more important than this but just an idea)

ironmig:
I noticed there are no changes to wamv.sdf in this PR. In the current version, there are some problems with the model. For example, the right camera is on the left side. The locations of the sensors as well (lidar, cameras) do not seems to be the same as we have in our TFs either. Will these be addressed in the sim?

whispercoros:
I am waiting for the new sensor mast to be installed before correcting all of the measurements for the sensors. I can temporarily move the right camera for this PR if need be. Did you manage get it to the proper location when you were using my fork last week? If so, post the pose here.

ironmig:
At least for X,Y,Z I think it should be the exact same as the TF, assuming they are both relative to the same point (measurement). I think waiting for the new mast is fine.

ironmig:
@whispercoros @mattlangford what's holding this up? Are all the requested changes addressed? It would be nice to have this in the main repo so I can use

mattlangford:
Yes, @whispercoros needs to add a way to launch the sim without spawning all the models so that our petty laptops can run the sim without cooking eggs

whispercoros:
My apologies @ironmig, I thought I had taken care of this PR. I have the changes sitting on the Samba server in MIL and will push them tomorrow.

whispercoros:
This can be merged once Semaphore finishes building it.

PR #86, opened 2016-10-15
DETECT DELIVER: Align target w/ lidar & search for color/shape

ironmig:
This is work from 10/10 - 10/14 on improving the mission for detect deliver. As per the RobotX manual, the mission will now only be given a color or shape (not both) to look for (the other parameter should be set to "ANY"). Also, the mission now has experimental functionality to align to the target using lidar. To achieve this, an service was added which can tranform a 2d camera point to an array of nearby 3d lidar points, as well as a normal to the plane formed by these points.


ironmig:
Just updated with changes from lake days 10/15-10/16. We're close to this working, but there are the following issues:

- The detect_deliver mission does not (always) align correctly to the normal found to the target plane.This is probably a fairly simple problem with the math of going from 
- The normal detected to the target plane may be incorrect due to it using liar points past the plane (greater z)
- No readjustment is done after navigating to the waypoint in the database, even though it may have drifted while moving
- No readjustment is done between shooting racquetballs, even though the target platform may move

ironmig:
Rebased. Build is failing because the tf2 sensor msgs package is not on semaphore. Perhaps this needs to be added to the install script @whispercoros?
`` sudo apt-get install ros-indigo-tf2-sensor-msgs ``

whispercoros:
@ironmig, it would make things simpler for you to add it and include it in this PR. I put the line in the right place, all you have to do is apply this diff:

    @@ -450,6 +450,9 @@ if ($INSTALL_NAV); then
     	apt-get source --compile -qq python-pyode
      	sudo dpkg -i python-pyode_*.deb
     
    +	# Message types
    +	sudo apt-get install -qq ros-indigo-tf2-sensor-msgs
    +
     	# Pulling large project files from Git-LFS
     	instlog "Pulling large files for Navigator"
     	cd $CATKIN_DIR/src/Navigator

Please not the added -qq to make the output less verbose and the fact that the install script uses tabs (yes, I am going to be particular on this).

ironmig:
Done

mission_control/navigator_missions/nav_missions/__detect_deliver.py
```diff
@@ -68,104 +48,122 @@ def get_waypoint(self):
...
-        #Allign perpendicular to the face of the target
-        print "Aligning perpendicular"
-
-
-    @txros.util.cancellableInlineCallbacks
-    def align_distance(self):
-        #Allign the correct distance from the target plane
-        print "Aligning distance"
-
+    def align_to_target(self):
```
mattlangford: Im going to wait to merge this until the issues we saw are resolved

mission_control/navigator_missions/nav_missions/__detect_deliver.py
```diff
@@ -4,8 +4,10 @@
 import std_srvs.srv
```
mattlangford: Download a linter to catch the PEP issues - there are a lot

sensor_control/vision/symbols/raylidar/CameraLidarTransformer.cpp
```diff
@@ -0,0 +1,211 @@
...
+      markers.markers.push_back(marker_point);
+
+      // Draw
+      if (int(point.x - 20) > 0 && int(point.x + 20) < camera_info.width &&
+          int(point.y - 20) > 0 && int(point.y + 20) < camera_info.height) {
+        cv::circle(debug_image, point, 3,
+                   cv::Scalar(0, 0, 255 * std::min(1.0, z.f / 20.0)), -1);
+      }
+#endif
+      double distance = sqrt(pow(point.x - req.point.x ,2) + pow(point.y - req.point.y,2) ); //Distance (2D) from request point to projected lidar point
```
mattlangford: Doesn't really matter here since speed doesn't seem to be a problem but something that will save some time would be not taking the square root and instead making `req.tolerance` in units of distance squared. Since this is getting run on each point, optimizations here effectively get multiplied.

sensor_control/vision/symbols/raylidar/CameraLidarTransformer.cpp
```diff
@@ -0,0 +1,211 @@
...
+       ++ii, jj += cloud_transformed.point_step) {
+    floatConverter x, y, z, i;
+    for (int kk = 0; kk < 4; ++kk) {
+      x.data[kk] = cloud_transformed.data[jj + kk];
+      y.data[kk] = cloud_transformed.data[jj + 4 + kk];
+      z.data[kk] = cloud_transformed.data[jj + 8 + kk];
+      i.data[kk] = cloud_transformed.data[jj + 16 + kk];
+    }
+    cam_model.fromCameraInfo(camera_info);
+    if (int(z.f) < 0 || int(z.f) > 30) continue;
```
mattlangford: Do you have to cast these arguments as ints?

utils/navigator_msgs/srv/CameraToLidarTransform.srv
```diff
@@ -0,0 +1,11 @@
+Header header #Stamp of time point was seen for tf
+geometry_msgs/Point point #X and Y of point in camera frame
+uint16 tolerance
```
mattlangford: What units are these? Meters? What if someone wants a half meter tolerance?

ironmig:
Requested documentation added. We get to merge (You promised!!!). We'll be adding stuff this weekend on top of this, but best to get this in dev now.

PR #87, opened 2016-10-16
PERCEPTION: ros camera stream class, tf changes, and fix to point cloud colorer

DSsoto:
ROSCameraStream is a templated class that greatly simplifies the acquisition of images from ros topics in c++ source code. It keeps a circular buffer of the most recent n frames on a topic and makes it easy to code perception algorithms that take into account temporal information. It also makes it easy to visualize and render objects on top of sequences of images as a whole. The camera frame objects that it stores contains pretty much all of the information you would need in a perception algorithm (for example: time stamp, tf frame id, camera intrinsics/extrinsics, distortion coefficients, etc.) It is also possible to request a camera frame object by index or by a time. See ros_camera_stream_demo.cpp for example usage.


mattlangford:
Never have I ever in allll my days seen code like this. Good job. Will merge once Semaphore is done.

PR #91, opened 2016-10-18
GUI: Dashboard for RQT

whispercoros:
This is the completed dashboard for RQT. It currently displays kill status, operating mode, battery voltage, system time, and the IP and network status of all network enabled devices on Navigator. It also has a control panel that implements most of the functionality of the X-Box controller (actual keyboard driving of the boat is in the works).

Test day was a surprisingly good time for edge case testing and I was able to take care of all of the ones I found while the boat was in a state of chaos, so it should be fairly stable. If any additional display or control features are needed, just ask. I made it very simple to implement new ones.


sensor_control/navigator_battery_monitor/nodes/navigator_battery_monitor.py
```diff
@@ -90,20 +89,30 @@ def publish_voltage(self, event):
...
+        self.voltage_check()
 
     def voltage_check(self):
         '''
         Checks the battery voltage and raises alarms when it's level drops below the set thresholds. This is intended to protect
         and extend the life of the batteries.
         '''
+
+        # An unknown battery status warning to inform users that thruster feedback is not being published
+        if (self.voltage is None):
```
mattlangford: This is a good idea - does the GUI alert us if this is the case?

sensor_control/navigator_battery_monitor/package.xml
```diff
@@ -14,8 +14,7 @@
   <run_depend>rospy</run_depend>
   <run_depend>std_msgs</run_depend>
   <run_depend>roboteq_msgs</run_depend>
-  <run_depend>sub8_alarm</run_depend>
+  <run_depend>navigator_alarm</run_depend>
```
mattlangford: Do Python scripts need this run_depend?

utils/navigator_gui/navigator_gui/dashboard.py
```diff
@@ -0,0 +1,486 @@
...
+        '''
+        # Attempts to read the battery voltage parameters (sets them to defaults if they have not been set)
+        self.battery_low_voltage = rospy.get_param("/battery_monitor/battery_low_voltage", 22.1)
+        self.battery_critical_voltage = rospy.get_param("/battery_monitor/battery_critical_voltage", 20.6)
+
+        rospy.Subscriber("/battery_monitor", Float32, self.cache_battery_voltage)
+        rospy.Subscriber("/clock", Clock, self.cache_system_time)
+
+        self.wrench_changer = rospy.ServiceProxy('/change_wrench', WrenchSelect)
+        self.station_holder = rospy.ServiceProxy('/lqrrt/station_hold', SetBool)
```
mattlangford: This has been changed - though I think I will add something back to support a service call to station hold. I might change it to an alarm

utils/navigator_gui/navigator_gui/dashboard.py
```diff
@@ -0,0 +1,486 @@
...
+            severity=0
+        )
+
+    def update_kill_status(self, alarm):
+        '''
+        Updates the kill status display when there is an update on the kill
+        alarm. Caches the last displayed kill status to avoid updating the
+        display with the same information twice.
+        '''
+        if (not self.system_time["is_timed_out"]):
```
mattlangford: Since a lot of your functions have this not timed out check, you could create a decorator to decorate all the functions with that would check for you. That would be a cool way to do it though this works exactly the same as that would and is easier.

utils/navigator_gui/navigator_gui/dashboard.py
```diff
@@ -0,0 +1,486 @@
...
+        # Schedules the next instance of this method with a QT timer
+        QtCore.QTimer.singleShot(500, self.monitor_operating_mode)
+
+    def update_operating_mode_status(self):
+        '''
+        Updates the displayed operating mode status text and color.
+        '''
+        if (self.operating_mode["received"] == "autonomous"):
+            self.operating_mode_status.setText("Autonomous")
+            self.operating_mode_frame.setStyleSheet(self.colors["green"])
```
mattlangford: I feel like colors here don't really make sense. It'd be cool to have to briefly flash a color whenever you change though...

mattlangford:
Looks mostly good. Before I merge though, we need to figure out a system to station hold the boat with the newest version of lqrrt.

sensor_control/navigator_battery_monitor/nodes/navigator_battery_monitor.py
whispercoros: The dashboard will count any message containing "None" towards the timeout counter for the battery voltage. Once this times out (after approximately 15s of None messages), the GUI will display "Unknown" as the battery voltage and change the color of it's display frame from green to red.

sensor_control/navigator_battery_monitor/package.xml
whispercoros: I was not sure about this. The battery monitor does import it, so it is technically a runtime dependency.

utils/navigator_gui/navigator_gui/dashboard.py
whispercoros: I was thinking about doing something like this as well. Now that there are multiple control devices (joystick, keyboard, joystick) it would be a good idea to implement a similar system to kills.

utils/navigator_gui/navigator_gui/dashboard.py
whispercoros: I will look in to this.

utils/navigator_gui/navigator_gui/dashboard.py
whispercoros: The colors are simply there to make the display glanceable. My thinking was that It is far easier to look at it and see a color for each item on the display than to read four words or values.

whispercoros:
@mattlangford, can you post the new station holding command here?

utils/navigator_gui/navigator_gui/dashboard.py
mattlangford: But blue doesn't mean anything at a glance does it?

mattlangford:
Create and alarm broadcaster and raise the `station_hold` alarm

utils/navigator_gui/navigator_gui/dashboard.py
whispercoros: It means Joystick in this context. You have to mentally build that association, but it becomes second nature.

whispercoros:
I just had a few more questions to make sure we are on the same page here. Does the alarm get cleared once the system enters station holding? Should the KillBroadcaster and SetBool parts of the station holding function be removed? What severity should be used (I went with a severity of 1, allowing something in the priority queue to supersede it while still keeping it fairly high)?

whispercoros:
Point noted. The language in the alarms page of the Sub8 wiki is a little confusing on the issue in that it makes it sound like the severity is used internally: "When many alarms must be handled, they will be handled as a priority queue..."

whispercoros:
This is good to merge once Semaphore checks out. I would like to get it on dev before Sunday so that we don't have to go through all of my relaunch shenanigans again.

mattlangford:
Semaphore is asleep :(
I agree getting this in dev is important

Issue #90, opened 2016-10-16
Object classifcation database should return the original data (Buoy.msg)


ironmig:
Currently, a service call to /database/single or /database/full returns data in the format of PerceptionObject.msg, which contains the objects id, name, size, and position. This removes some of the data of the original Buoy.msg sent by object detection, such as a normal and the 3D points of the object. For use in missions and for less confusion, perhaps this should just be one message format?

mattlangford:
I think the plan right now is for @maverick7170 (Ira) to get a database full of automatically classified objects for this weekend. I'm going to assign him on this too - make sure he knows if you want more information in the message then what currently exists.

maverick7170:
Yes, I'll have this ready for this weekend!

Issue #82, opened 2016-10-12
Create constants for object database keys


ironmig:
So missions can know what object to search the database for, it would be helpful to define constants for the keys to use in the message file. For example

string SHOOTER=detect deliver target
string DOCK=find and dock

mattlangford:
Whoops wrong button

PR #92, opened 2016-10-19
MISSION SYSTEM: Clean database/object classifcation

ironmig:
Combine Buoy.msg (non descriptive) and PerceptionObject. Now there is PerceptionObject, which are not identified and IdentifiedPerceptionObject, which contains a string for its name and a copy of the original object

Allows missions which call the database to receive the original data of the object, including the points.

Addresses both #90 and #82.

I know multiple systems might be affected by this change, and I did my best to update them to use the new messages. I may have missed something (especially in the python code) so please look through before merge. @tessbianchi, are you able to test the object classifcation/database stuff to see if I broke it before lakeday? @maverick7170, can you test your code with these message changes as well?


maverick7170:
@ironmig I'm actually going to be rolling all of this code back into the original lidar node! @tessbianchi  has been doing a lot of extra work to get my code to fit in with out parts of the system. I think the best option is for me to roll her code back into mine and maintain it, freeing @tessbianchi  to focus on other things! 
1) Can we combine IdentifiedPerceptionObject and PerceptionObject into a single message?
2) When you make your service request, are there any other features/info you want?

ironmig:
Maybe we should combine IdenfiedPerceptionObject and PerceptionObject into one by just adding a string for the type of object it is, which will be set to UNKOWN until it is classified. I do think having some information about the orientation of the object is important, such as the normal you had at lakeday.

ironmig:
That will keep things simple. I'll add a commit with this change.

tessbianchi:
Hi Ira!

I don't think rolling back all of the code is a good idea, because I implement persistant object ids, which is really useful. I think it would be best to put the code on top of the code that I have  written. 

Also, I have a few topics that I want from the database.

'/database/object_found', PerceptionObject = This publishes an object every time a new object a found, but it doesn't republish the same objects over and over again
'/database/objects', PerceptionObjects = This publishes a list of objects every 2 or three seconds.

'/database/single', ObjectDBSingleQuery = This takes a name as a string and returns a the perception object in the database
'/database/full', ObjectDBFullQuery = This should return a list of the full database

ironmig:
@rleyva Is planning on changing the message to publish a PointCloud2, perhaps in addition to the geometry points. We should merge this (ideally after testing with a bag) to allow Ralph to change the message and resolve the conflicts.

PR #97, opened 2016-10-21
Lidar

maverick7170:
These commits include the latest lidar code with several bug fixes and Tess's code rolled back into the lidar node.

Start the code with rosrun:
rosrun navigator_lidar_oa lidar

It currently publishes topics with my name in front (this should change...):
ira_ogrid
ira_markers
ira_database

You can interact with the database a few different ways. To get objects:
rosservice call /ira_database "name: 'XXX'" where XXX is shooter, scan_the_code, all, etc...

To update information about objects:
rosservice call /ira_database "cmd: 'XXX=YYY'" where XXX is the numeric id of the object and YYY is the value to change. 

For example:
rosservice call /ira_database "cmd: '1=start_gate'"
rosservice call /ira_database "cmd: '3=scan_the_code'"

If you list numeric values, you can change the RGB of an object in the database:
rosservice call /ira_database "cmd: '5=255,0,0'"

The code doesn't include point cloud coloring yet so we can manually set the color for now.


sensor_control/navigator_lidar_oa/src/VolumeClassifier.h
```diff
@@ -0,0 +1,21 @@
...
+	if (object.name == navigator_msgs::PerceptionObject::START_GATE_BUOY || object.name == navigator_msgs::PerceptionObject::SCAN_THE_CODE) {
+		return;
+	}
+	//if (object.name == navigator_msgs::PerceptionObject::BUOY || object.name == navigator_msgs::PerceptionObject::UNKNOWN) {
+		if (object.scale.x > 4 && object.scale.y > 4 && object.scale.z > 2 && object.strikesPersist.size() > 600 ) {
+			object.name = navigator_msgs::PerceptionObject::DETECT_DELIVER_PLATFORM;
+		} else if (object.scale.x > 1.5 && object.scale.y > 1.5 && object.scale.z > 1.5 && object.strikesPersist.size() > 300) {
+			object.name = navigator_msgs::PerceptionObject::TOTEM;
+		} else if (false) {
+			object.name = navigator_msgs::PerceptionObject::SCAN_THE_CODE;
```
tessbianchi: What are the sizing constraints going to be here?

sensor_control/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -66,24 +65,11 @@ uf_common::PoseTwistStamped waypoint_enu,carrot_enu;
...
-	ROS_INFO("LIDAR: Following new goal from action server!");
-	waypoint_enu.posetwist = goal->posetwist; 
-}
+//Do these come from a rosparam or service call?
+Eigen::Vector2d BOUNDARY_CORNER_1 (30, 10);
+Eigen::Vector2d BOUNDARY_CORNER_2 (30, 120);
+Eigen::Vector2d BOUNDARY_CORNER_3 (140, 120);
+Eigen::Vector2d BOUNDARY_CORNER_4 (140, 10);
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
tessbianchi: I didn't have this implement yet, but it would be nice if this could get the data directly from the rosservice: /get_bounds

sensor_control/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -386,163 +388,30 @@ int main(int argc, char* argv[])
...
 	//Subscribe to odom and the velodyne
 	ros::Subscriber sub1 = nh.subscribe("/velodyne_points", 1, cb_velodyne);
 	ros::Subscriber sub2 = nh.subscribe("/odom", 1, cb_odom);
 
 	//Publish occupancy grid and visualization markers
-	pubGrid = nh.advertise<nav_msgs::OccupancyGrid>("ogrid_batcave",10);
-	pubMarkers = nh.advertise<visualization_msgs::MarkerArray>("/unclassified/objects/markers",10);
-	pubBuoys = nh.advertise<navigator_msgs::PerceptionObjects>("/unclassified/objects",10);
+	pubGrid = nh.advertise<nav_msgs::OccupancyGrid>("ira_ogrid",10);
+	pubMarkers = nh.advertise<visualization_msgs::MarkerArray>("ira_markers",10);
```
tessbianchi: This should be 

/unclassified/markers

sensor_control/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -367,6 +305,71 @@ void cb_odom(const nav_msgs::OdometryConstPtr &odom) {
...
+//Handle DB Requests
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+bool objectRequest(navigator_msgs::ObjectDBQuery::Request  &req, navigator_msgs::ObjectDBQuery::Response &res) 
+{
+	ROS_INFO_STREAM("LIDAR | DB request with name " << req.name << " and command " << req.cmd);
+	res.found = false;
+	res.objects.clear();
+
+	//Did we get a command?
+	//Each command needs an id and value seperated by equals sign, this is ugly code...
```
tessbianchi: Does this support a way to get the whole database at once?

sensor_control/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -386,163 +388,30 @@ int main(int argc, char* argv[])
...
+	pubGrid = nh.advertise<nav_msgs::OccupancyGrid>("ira_ogrid",10);
+	pubMarkers = nh.advertise<visualization_msgs::MarkerArray>("ira_markers",10);
+	pubCloudPersist = nh.advertise<sensor_msgs::PointCloud>("ira_persist",1);
+	pubCloudFrame = nh.advertise<sensor_msgs::PointCloud>("ira_frame",1);
+	pubCloudPCL = nh.advertise<sensor_msgs::PointCloud>("ira_pclcloud",1);
 
-	//Publish waypoints to controller
-	//pubTrajectory = nh.advertise<uf_common::PoseTwistStamped>("trajectory", 1);
-    //pubWaypoint = nh.advertise<PoseStamped>("waypoint", 1); //Do we need this?
+	//Service for object request
```
tessbianchi: This should be /database/single


I also need the functionality to get the whole database at once. Is this possible to add?

sensor_control/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -386,163 +388,30 @@ int main(int argc, char* argv[])
...
+	ros::ServiceServer service = nh.advertiseService("ira_database", objectRequest);
 
-	// pubMarkersSmall = nh.advertise<visualization_msgs::MarkerArray>("/vision/objects_unclassified",10);
 	//Give control to ROS
 	ros::spin();
```
tessbianchi: I also need these topics to be published.

'/database/object_found' returns a PerceptionObject = This publishes an object every time a new object a found, but it doesn't republish the same objects over and over again
'/database/objects', returns PerceptionObjects = This publishes a list of objects every 2 or three seconds.

sensor_control/navigator_lidar_oa/src/VolumeClassifier.h
maverick7170: Andy discussed this morning about verifying the height of the competition scan the code structure and making ours match. We'll be able to pick bounds then! Until then, we can set the name manually.

sensor_control/navigator_lidar_oa/src/lidar.cpp
maverick7170: Sure! Is there a description of the bounds service or just grab it from the source code?

sensor_control/navigator_lidar_oa/src/lidar.cpp
maverick7170: Using the name "all" will just return everything.

sensor_control/navigator_lidar_oa/src/lidar.cpp
maverick7170: Sure!

sensor_control/navigator_lidar_oa/src/lidar.cpp
maverick7170: Who will be using the published objects vs the service calls?

mattlangford:
The build failed due to `ObjectDBSingleQuery` having been removed. I think having just one object return a list of objects would be a good way to do things. 
Two things:

- I don't think the missions should rely on their own service proxies for something so common, ie. they should use a navigator object service proxy (that don't exist yet) to get database objects.
- That functionality hasn't been added yet so I'm not sure what we should do for this PR. Manually having everyone correct their missions would get messy since people have likely made changes to their missions in their own branches. We might just ignore the build error and have everyone fix their errors in upcoming PR's (or better yet use the non existent navigator singleton `database_query` tx function that will exist by tonight in `dev`).

maverick7170:
Matt, I didn't think about other code depending on ObjectDBSingleQuery... Would it be easier to just to add those files back on my fork and create a new PR? Essentially, what can I do to help fix things?

mattlangford:
I'm not sure what the best way to work it out is. I think if possible we should keep it removed since I think one service makes things a lot cleaner. I think I can pull your PR and make the appropriate changes but that would break things down the road. Later tonight I will get with mission people and have them all make changes in their code to not using the soon to be depreciated service type.

So I would say don't worry about it right now (unless Tess really needs that service type back in) - I will figure a way to work around it.

maverick7170:
Ok, sounds good!

utils/navigator_msgs/msg/PerceptionObject.msg
```diff
@@ -8,16 +8,21 @@ string IDENTIFY_AND_DOCK=dock
...
 
 uint32 id                         # If we want to label each buoy and keep track of it
 uint8 confidence                  # 0-255 how confident are we that this is a buoy
 
-float64 height                    # Height and width for the detected object
-float64 width                     # Can be used for detecting different objects
-float64 depth					  # Third dimension in x,y,z...
+float64 zHeight                    # Height and width for the detected object
+float64 xWidth                     # Can be used for detecting different objects
+float64 yDepth					  # Third dimension in x,y,z...
```
ironmig: Has @rleyva talked to you about including a PointCloud2 in this? That could replace intensity and points.

Also, what do you think about combining zHeight, xWidth, and yDepth into one object such as a geometry_msgs/Vector3 to represent the size to keep the message clean?

utils/navigator_msgs/srv/ObjectDBQuery.srv
```diff
@@ -0,0 +1,5 @@
+string name
+string cmd
```
ironmig: Could you add some comments into this service file about its usage? For example, what does cmd do?

ironmig:
Just for reference this addresses #90 and #82.

sensor_control/navigator_lidar_oa/src/lidar.cpp
maverick7170: Thanks! I'll get this added.

utils/navigator_msgs/msg/PerceptionObject.msg
maverick7170: @rleyva has mentioned this, but I haven't seen code or don't know where to pull it from. I don't really have a preference for how the data is stored but does PointCloud2 get us something we need? 

Creating geometry_msgs/Vector3 works!

sensor_control/navigator_lidar_oa/src/VolumeClassifier.h
tessbianchi: Ok sounds good!

sensor_control/navigator_lidar_oa/src/lidar.cpp
tessbianchi: ok great!

utils/navigator_msgs/msg/PerceptionObject.msg
jpanikulam: PointCloud2 can be directly deserialized to a PCL cloud type, with no need for custom code

sensor_control/navigator_lidar_oa/src/lidar.cpp
tessbianchi: Ok, that is what it was before. As long as it starts with /database/ then that's fine!

sensor_control/navigator_lidar_oa/src/lidar.cpp
tessbianchi: The mission system needs to get notified when a new object is added

Also, it will just be helpful in general to have a constantly updating list that any node can use. Right now I use it to tell Matt what the closest unknown object is so that we can classify 

It is possible to use the service call over and over again, but I think this will be a lot of extra work for the other nodes to implement.

sensor_control/navigator_lidar_oa/src/lidar.cpp
mattlangford: Make a service call to that service with the data field being "enu" and it'll give a response with 4 enu points

maverick7170:
Closing this PR to start from Matt's copy of the lidar code used on the boat last Sunday.

mattlangford:
I will be pushing by Wednesday night (due to a couple exams and projects coming up) - unless someone wants to get all the changes off the boat and commit them appropriately.

PR #105, opened 2016-10-27
SCRIPTS: Bag backup script

whispercoros:
This bag backup script is similar to the one used on Sub8, but it has been changed to use locally mounted directories that can be passed as arguments from /etc/rc.local or the command line. I have added the line in the comments section of this file to /etc/rc.local with the variables set in properly. Testing was successful.

In the future, follow the instructions in the comments section of the script after test days to synchronize bag files to the external drive. Once the drive is in MIL, it would also be courteous to upload the files to the Samba server for everyone.

### DIFF
--- a/scripts/bag_backup.sh
+++ b/scripts/bag_backup.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+# This script can be used to sync all new bags in the bags directory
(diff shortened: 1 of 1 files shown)


whispercoros:
I should also point out that I manually made these changes on the boat. Once this is merged to dev, they should be replaced by pulling dev before the next test day.

PR #106, opened 2016-10-28
Lidar

maverick7170:
PR for new lidar changes shown in the meeting today (interactive markers, avoiding water reflection, new volume estimates). Will write up an explanation and more comments to code soon!


perception/navigator_lidar_oa/src/ConnectedComponents.h
```diff
@@ -83,6 +83,9 @@ struct objectMessage
 	uint32_t pclInliers = 0;
 	geometry_msgs::Vector3 normal;
 	float minHeightFromLidar,maxHeightFromLidar;
+	bool current = false;
+	bool locked = false;
```
ironmig: One thing I would like is to be able to interact with these members from ros messages / the database service besides just rviz. So perhaps we should just add locked and real to the PerceptionObject message, and some commands to the database service for setting these/inserting new ones. This way we could write another program which, for example, inserts GPS waypoints with given object types, or presets some default "fake" object locations.

perception/navigator_lidar_oa/src/ConnectedComponents.h
```diff
@@ -182,10 +185,9 @@ std::vector< std::vector<int> > ConnectedComponents(OccupancyGrid &ogrid, std::v
 	}
 
 	//Re-organize obstacles, how many connections to count as an obstacle?
-	int newId = 0;
 	objects.clear();
 	for (auto ii : mapObjects)  {
-		if (ii.second.strikesPersist.size() >= 10) {
```
ironmig: This might not be needed for this PR, but eventually magic numbers like this should be setup as rosparams, so they can easily be changed /viewed at runtime. This is as easy as:
```
 nh.param<int>("strikes_persist",STRIKE_PERSIST_VARIABLE, 10);
```
Where the string is the key we set the param with, the middle term is the variable to set by reference, and the third is a default. The template can be int,double,string etc.

ironmig: Minor gripe, but be careful committing file mode changes. I think you accidentally committed changes to group permissions, which could be an issue if we ever try to edit these files from another user account

perception/navigator_lidar_oa/src/ConnectedComponents.h
maverick7170: Agreed and this was actually the original plan. I actually wrote this rosservice code before but it isn't part of this PR. We can certainly add it back.

perception/navigator_lidar_oa/src/ConnectedComponents.h
maverick7170: Good with me. How often should a node check for possible changes from a rosparam?

maverick7170: Hmm.... I didn't actively change any file modes on my end, so is this a result of the git merge moving files? In any case, easy to fix and update the PR!

PR #108, opened 2016-10-28
SCRIPTS: Minor changes to the install script

whispercoros:
This sets up ros_connect to work with the secondary computer on NaviGator and installs some missing git repositories. It also includes a few bug fixes for the dashboard that are too small for a separate PR.


mattlangford:
Does the google maps rviz stuff work yet? Im excited for that.

Looks like harmless changes, will merge after we merge whenever the build passes.

whispercoros:
Not quite yet, just wanted to get it in the script while I was working on it. I actually do need to add the lines to build lqRRT, so merge after that commit builds. Thanks for the reminder!

PR #109, opened 2016-10-29
Lidar: new readme with guide on how to use node

maverick7170:
Mostly cosmetic changes but wanted to get the readme in a PR for explanations of how to use the code. Here is a copy of the readme:

##Quick Overview

**Required Interfaces**

	- /velodyne_points
	- /odom
	- /get_bounds

**Provided Interfaces**

	- /ogrid
	- /info_markers
	- /database/objects
	- /database/requests
	- /unclassified_markers

**Interfaces for Debugging**

	- /ira_persist
	- /ira_frame
	- /ira_pclcloud
  

##Quick Start Guide

**Starting the node**

	- rosrun navigator_lidar_oa lidar

**Requesting objects from the database**

The service is name /database/requests with the input parameter name and will return the PerceptionObject message. Here are some example calls:

	- rosservice call /database/requests "name: 'all'" 
	- rosservice call /database/requests "name: 'shooter'"
	- rosservice call /database/requests "name: 'totem'"
	- rosservice call /database/requests "name: 'scan_the_code'"
	- rosservice call /database/requests "name: 'start_gate'" 
	- rosservice call /database/requests "name: 'buoy'"

**Setting ROI objects from the database**

ROI (region of interest) allows us to set fake markers in the database as estimates about where structures are located in the ENU world frame. This can either be done through rosservice with the cmd input or manually in RVIZ. The fake markers include: **BuoyField**, **CoralSurvey**, **FindBreak**, **AcousticPinger**, **Shooter**, **Scan\_The_Code**. (These can be changed/standardized later...). The syntax requires the name of the fake marker and the **x,y location in the ENU** frame after the = sign. Example calls:

	- rosservice call /database/requests "cmd: 'Shooter=-27,-69'"
	- rosservice call /database/requests "cmd: 'CoralSurvey=50.5,71.024'"

**Interactive Markers in RVIZ**

Subscribing to **/unclassified_markers** will allow RVIZ to setup interactive markers for objects found by the lidar. Right clicking on a marker will bring up the following menu:

	- Type
	- Lock
	- Color

Type allows you to change the object classification, lock will prevent the node from classifying this particular object in code, and color sets the rgb value of the object.
Note that an unknown color defaults to 0,0,0 for rgb which is black, but we do have black buoys. So a black buoy is really 0.01,0.01,0.01 in rgb. (This can be addressed later!)

The ROI fake markers will also show up in RVIZ at the origin. Simply drag these to the necessary locations.

##Key parameters for the code
These are the knobs to turn to control the behavior of the ogrid  (these will eventually become rosparams for easy editing). Most units are marked in the variable name but all variables use meters and degrees.

	const double MAP_SIZE_METERS = 1500;
	const double ROI_SIZE_METERS = 201;
	const double VOXEL_SIZE_METERS = 0.30;
	const int MIN_HITS_FOR_OCCUPANCY = 25; 
	const int MAX_HITS_IN_CELL = 125; 
	const double MAXIMUM_Z_BELOW_LIDAR = 2; 
	const double MAXIMUM_Z_ABOVE_LIDAR = 2.5;
	const double MAX_ROLL_PITCH_ANGLE_DEG = 5.3;
	const double LIDAR_VIEW_ANGLE_DEG = 160;
	const double LIDAR_VIEW_DISTANCE_METERS = 80;
	const double LIDAR_MIN_VIEW_DISTANCE_METERS = 5.5;
	const int MIN_LIDAR_POINTS_FOR_OCCUPANCY = 10;
	const double MIN_OBJECT_HEIGHT_METERS = 0.075;

### DIFF
--- a/perception/navigator_lidar_oa/src/OccupancyGrid.h
+++ b/perception/navigator_lidar_oa/src/OccupancyGrid.h
@@ -122,11 +122,13 @@ class OccupancyGrid
 	    /// \param ?
 	    /// \param ?
 	    ////////////////////////////////////////////////////////////
-		void setLidarParams(const double LIDAR_VIEW_ANGLE_DEG, const double LIDAR_VIEW_DISTANCE_METERS, const double LIDAR_MIN_VIEW_DISTANCE)
+		void setLidarParams(const double LIDAR_VIEW_ANGLE_DEG, const double LIDAR_VIEW_DISTANCE_METERS, const double LIDAR_MIN_VIEW_DISTANCE, const int MIN_LIDAR_POINTS_FOR_OCCUPANCY, const double MIN_OBJECT_HEIGHT_METERS)
 		{
 			lidarViewAngle = LIDAR_VIEW_ANGLE_DEG;
 			lidarViewDistance = LIDAR_VIEW_DISTANCE_METERS;
-                        lidarMinViewDistance = LIDAR_MIN_VIEW_DISTANCE;
+            lidarMinViewDistance = LIDAR_MIN_VIEW_DISTANCE;
+            lidarMinPoints = MIN_LIDAR_POINTS_FOR_OCCUPANCY;
+            objectMinHeight = MIN_OBJECT_HEIGHT_METERS;
 		}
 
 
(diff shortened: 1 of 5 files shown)


viron11111:
Don't forget about the docking challenge.

-Andy

perception/navigator_lidar_oa/src/OccupancyGrid.h
```diff
@@ -122,11 +122,13 @@ class OccupancyGrid
...
 	    ////////////////////////////////////////////////////////////
-		void setLidarParams(const double LIDAR_VIEW_ANGLE_DEG, const double LIDAR_VIEW_DISTANCE_METERS, const double LIDAR_MIN_VIEW_DISTANCE)
+		void setLidarParams(const double LIDAR_VIEW_ANGLE_DEG, const double LIDAR_VIEW_DISTANCE_METERS, const double LIDAR_MIN_VIEW_DISTANCE, const int MIN_LIDAR_POINTS_FOR_OCCUPANCY, const double MIN_OBJECT_HEIGHT_METERS)
 		{
 			lidarViewAngle = LIDAR_VIEW_ANGLE_DEG;
 			lidarViewDistance = LIDAR_VIEW_DISTANCE_METERS;
-                        lidarMinViewDistance = LIDAR_MIN_VIEW_DISTANCE;
+            lidarMinViewDistance = LIDAR_MIN_VIEW_DISTANCE;
+            lidarMinPoints = MIN_LIDAR_POINTS_FOR_OCCUPANCY;
+            objectMinHeight = MIN_OBJECT_HEIGHT_METERS;
```
DSsoto: The alignment here is off

perception/navigator_lidar_oa/src/lidar.cpp
```diff
@@ -255,16 +276,8 @@ void cb_velodyne(const sensor_msgs::PointCloud2ConstPtr &pcloud)
...
-	  	visualization_msgs::InteractiveMarkerControl control;
-	  	int_marker.header.frame_id = "enu";
-	  	int_marker.scale = 1;
-	  	int_marker.name = to_string(obj.id);
-		control.interaction_mode = visualization_msgs::InteractiveMarkerControl::BUTTON;
-	  	control.always_visible = true;
-	  	control.name = to_string(obj.id);
-
+		//Create Interactive marker
+		auto int_marker = CreateInteractiveMarker(to_string(obj.id));
```
DSsoto: Very nice! This will be very useful

PR #111, opened 2016-10-30
LAUNCH: Add lqrrt,  vision nodes, lidar oa / database to master launch

ironmig:
When me and @whispercoros were testing the sim, we ran into issues with GNC because essential components are not launched by default. I think its a good time to add everything that is needed for the robot to operate autonomous to the master launch file (or one of its included files) so we don't need to remember to launch everything. I went ahead and added everything I could think of instead of just making an issue. Let me know if you disagree about where something should go or w/e

### DIFF
--- a/mission_control/navigator_launch/launch/gnc.launch
+++ b/mission_control/navigator_launch/launch/gnc.launch
@@ -24,20 +24,9 @@
         <param name="torque_scale" value="500" type="double"/>
     </node>
 
+    <include file="$(find navigator_launch)/launch/subsystems/lqrrt_node.launch"/>
+    <node name"lidar_ogrid_database" pkg="navigator_lidar_oa" type="lidar" />
     <!-- Legacy - should be phased out -->
     <node pkg="kill_handling" type="kill_master" name="kill_master"/>
-    <node pkg="c3_trajectory_generator" type="c3_trajectory_generator" name="c3_trajectory_generator" output="screen">
-        <rosparam>
-            <!-- Degraded trajectory -->
-            fixed_frame: /enu
-            body_frame: /base_link
-            vmin_b: [-0.68, -0.45, -1, -1, -1, -0.19]
-            vmax_b: [  1.1,  0.45,  1,  1,  1,  0.19]
-            amin_b: [-0.34, -0.15, -1, -1, -1, -0.04]
-            amax_b: [ 0.53,  0.15,  1,  1,  1,  0.04]
-            arevoffset_b: [0.0, 0, 0]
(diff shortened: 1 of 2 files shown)


mission_control/navigator_launch/launch/master_boat.launch
```diff
@@ -2,5 +2,6 @@
     <include file="$(find navigator_launch)/launch/subsystems/hardware_drivers.launch"/>
     <include file="$(find navigator_launch)/launch/gnc.launch"/>
     <include file="$(find navigator_launch)/launch/subsystems/sensors.launch"/>
+    <include file="$(find navigator_launch)/launch/perception.launch" />
     <include file="$(find navigator_alarm)/launch/alarm.launch"/>
```
DSsoto: @mattlangford is there a reason for alarm.launch not being in the navigator_launch pkg?

ironmig:
Lqrrt has its own launch, which I included from gnc.launch

mission_control/navigator_launch/launch/gnc.launch
```diff
@@ -24,20 +24,9 @@
         <param name="torque_scale" value="500" type="double"/>
     </node>
 
+    <include file="$(find navigator_launch)/launch/subsystems/lqrrt_node.launch"/>
```
mattlangford: I told someone this but lqrrt is still in a place where there is some debugging that goes on at lake day and sometimes it needs to be restarted and it's output need to be seen. When you launch with the entire stack the output will be mixed in with the output from everything which makes it hard to work with and restarting isn't easy.

mission_control/navigator_launch/launch/gnc.launch
```diff
@@ -24,20 +24,9 @@
...
+    <node name"lidar_ogrid_database" pkg="navigator_lidar_oa" type="lidar" />
```
mattlangford: `name=`
I think launching this here is a good idea

mission_control/navigator_launch/launch/master_boat.launch
mattlangford: Nope, it probably should be though.

ironmig:
I'll add the machine tags once I've successfully tested it (@whispercoros Tuesday). I'll take out lqrrt then too.

mission_control/navigator_launch/launch/gnc.launch
whispercoros: @mattlangford, I think it is important for either lqRRT or C3 (I would really prefer lqRRT) to come up with the system in the sim. How would you feel about a lake_day argument that skips launching it?

ironmig:
Added to #114

PR #114, opened 2016-11-01
DETECT DELIVER: Lakeday bug fixes and modularity for future missions

ironmig:
- Fixed bug causing detect deliver alignment normal to be wacky when using front camera. It worked with bag testing, so we should now be able to run the mission with either camera with only changing the launch files and it will "just work" \#ros
- Published topics for symbol vision node can now be defined in launch file, so in the future multiple versions of the node can be run for detect deliver, identify symbols and dock, and coral survey
- Switched the launch files to use the front stereo (left) for detect deliver , while side camera is still terrible
- You can now specify a resolution to resize the camera from to for vision processing in the symbols vision node, so it can still run reasonably fast on high res cameras (like the front ones)
- Add visualization messages to detect deliver, so you can see in rviz where it thinks the target is and where it wants to move to align with it
- General code cleaning in detect deliver mission, such as all magic numbers defined as class constants at the top of the file


mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -11,22 +11,28 @@
...
 from twisted.internet import defer
 from image_geometry import PinholeCameraModel
+from visualization_msgs.msg import Marker,MarkerArray
 import navigator_tools
 
 
 class DetectDeliverMission:
     # Note, this will be changed when the shooter switches to actionlib
-    resp = None
+    shoot_distance_meters = 5
```
mattlangford: Is there a reason you don't put these in the init function? If I was looking at code and was looking for the value `self.circle_radius` which is used later in the code, I'd control-f for `self.circle_radius` which wouldn't exist.

This space is used for static class variables more so than just general parameters

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -106,32 +135,51 @@ def fire_execute_callback(self, goal):
...
         self.stop_actions()
-        self.motor_controller.setMotor1(-req.feeder)
-        self.motor_controller.setMotor2(req.shooter)
-        rospy.loginfo("manual set")
-        return True
+        res = ShooterManualResponse()
+        res.success = True
+        return res
+
+    def isJoyShoot(self,data): #Shoot if right trigger is held down
```
mattlangford: Mixing_stylesAgain

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -106,32 +135,51 @@ def fire_execute_callback(self, goal):
...
+      return data.axes[5] < -0.9
+    def isJoyLoad(self,data): #load if left bumper is pressed
+      return data.buttons[4] == 1
+    def isJoyCancel(self,data): #cancel if right bumper is pressed
+      return data.buttons[5] == 1
+    def joy_callback(self,data):
+        joy_shoot = self.isJoyShoot(data)
```
mattlangford: Is there a motivation to doing it like this with the one line return functions?

ironmig: I wanted to make it easy to change what button/axis does what.

mattlangford:
Let me know when you want to merge this

You should make the launch changes in here before that though

ironmig:
Launch file changes are in there, except alarms and llrqt which you wanted to do. Merge is a go

mattlangford:
What about `/is_simulation`?

Issue #117, opened 2016-11-03
Navigator singelton should have methods for setting mission parameters


ironmig:
It would be nice if mission-specific parameters could be set through the navigator singleton with methods similar to what we have implemented the vision services. A yaml file would list all the mission parameters and their valid values. 

Ex:
```
detect_deliver_shape:
  param: /mission/detect_deliver/Color
  args: ['GREEN','BLUE','RED']
```

Then in a mission (like scan the code):

```
if scan_the_code_color[2] == "RED"
  #Throws error if color is invalid, warnings if already set, etc
  navigator.set_mission_param('detect_deliver_shape','RED')

```

I'll work on this. Not sure how this will work into mission planer system...

mattlangford:
You have the power
But I will do it if you'd like

PR #118, opened 2016-11-04
MISSIONS: navigator object provides methods for missions to safely get yaml defined paramters

ironmig:
Addresses #117 

Adds an object to the navigator mission object similar to the vision services dictionary, where missions can get a param by an index, which gets the corresponding param in the yaml file, checking that its value is one of the valid values. Also has the added benefit of forcing us to document which params we are going to use for missions. The mission planner (what is the status on that anyways?) could use this same yaml file or the parsed dictionary itself to make decisions about when to run missions.

### DIFF
--- a/mission_control/navigator_missions/navigator_singleton/mission_params.yaml
+++ b/mission_control/navigator_missions/navigator_singleton/mission_params.yaml
@@ -1,3 +1,10 @@
+explorer_next:
+  param: /mission/explorer/next
+  options: [ALL,BuoyField,CoralSurvey,FindBreak,
+    AcousticPinger,Shooter,Scan_The_Code,Gate_1,Gate_2,Gate_3,Dock]
+  description: The next unknown object to search for when no missions are running, defaults to furthest away (ALL)
+  default: ALL
+
 detect_deliver_color:
   param: /mission/detect_deliver/Color
   options: [RED, GREEN, BLUE]
@@ -23,7 +30,7 @@ scan_the_code_color3:
   options: [RED,GREEN,YELLOW,BLUE]
   description: The third color in the 3 color sequence of scan the code
 
-coral_surey_shape2:
+coral_surey_shape1:
   param: /mission/coral_survey/shape1
   options: [CIRCLE, TRIANGLE, CROSS]
   description: The shape identified in quadrant I of the Underwater Shape Identification challenge
--- a/mission_control/navigator_missions/navigator_singleton/mission_params.yaml
+++ b/mission_control/navigator_missions/navigator_singleton/mission_params.yaml
@@ -1,7 +1,81 @@
 detect_deliver_color:
   param: /mission/detect_deliver/Color
   options: [RED, GREEN, BLUE]
+  description: The color of the symbol to shoot at in detect deliver
 
 detect_deliver_shape:
   param: /mission/detect_deliver/Shape
   options: [CIRCLE, TRIANGLE, CROSS]
+  description: The shape of the symbol to shoot at in detect deliver
+
+scan_the_code_color1:
+  param: /mission/scan_the_code/color1
+  options: [RED,GREEN,YELLOW,BLUE]
+  description: The first color in the 3 color sequence of scan the code
+
+scan_the_code_color2:
+  param: /mission/scan_the_code/color2
+  options: [RED,GREEN,YELLOW,BLUE]
+  description: The second color in the 3 color sequence of scan the code
+
+scan_the_code_color3:
+  param: /mission/scan_the_code/color3
+  options: [RED,GREEN,YELLOW,BLUE]
+  description: The third color in the 3 color sequence of scan the code
+
+coral_surey_shape2:
+  param: /mission/coral_survey/shape1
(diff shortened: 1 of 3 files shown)


mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -221,6 +238,57 @@ def get_response(self, **kwargs):
 
         return self.client(s_req)
 
+class MissionParam(object):
```
mattlangford: Put all these as 4 spaces/tab please

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -221,6 +238,57 @@ def get_response(self, **kwargs):
...
+    def __init__(self, nh, param, options):
+        self.nh = nh
+        self.param = param
+        self.options = options
+
+    @util.cancellableInlineCallbacks
+    def get(self):
+        # Returns deferred object, make sure to yield on this (same for below)
+        if not (yield self.exists()):
+          raise Exception("Mission Param {} not yet set".format(self.param))
```
mattlangford: Maybe have default values specified in the yaml that are loaded if a param isn't found?

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -221,6 +238,57 @@ def get_response(self, **kwargs):
...
+        value = yield self.nh.get_param(self.param)
+        if not self._valid(value):
+          raise Exception("Value {} is invalid for param {}. Valid values: {}".format(value, self.param, self.options))
```
mattlangford: Or default if it's not valid

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -221,6 +238,57 @@ def get_response(self, **kwargs):
...
+        else:
+          defer.returnValue(value)
+
+    @util.cancellableInlineCallbacks
+    def exists(self):
```
mattlangford: You can just make this a normal function as follows (for cleanliness sake)
```python
# no decorator
def exists(self):
    return self.nh.has_param(self.param)
```
This will return a defer object that can be yeileded on. So calling the function is the same with less clutter in the function: `yield self.exists()`

mission_control/navigator_missions/navigator_singleton/mission_params.yaml
```diff
@@ -0,0 +1,7 @@
+detect_deliver_color:
+  param: /mission/detect_deliver/Color
+  options: [RED, GREEN, BLUE]
```
mattlangford: Do these need to be strings?

ironmig: they can be, but yaml makes them strings automatically

mission_control/navigator_missions/navigator_singleton/navigator.py
ironmig: I feel like for most of these if it isn't set, an error should be thrown.

ironmig:
I just thought this would be a useful abstractions for the missions, and would provide a place to document what params a mission uses and what values they can be. For example, Tess was setting "circle" instead of "CIRCLE" in Scan the Code. Detect deliver would then run with an invalid parameter, and just return "no shape found" forever without any obvious indication that the parameter was wrong. Other programs, like the GUI could also use this to abstract this information to a central place.

mattlangford:
Okay, we can merge this after the dev master merge after the lake day merge

mission_control/navigator_missions/navigator_singleton/navigator.py
mattlangford: I feel like you should tell the mission it doesn't exist and let them deal with it

ironmig:
Going to add in more params tomorrow. Here's a list (comment if you think of any more)

Scan the code colors ("RGB", "YRB" etc)

Identify and dock 
- left "GREEN" "CIRCLE"
- middle "RED" "TRIANGLE"
- right "BLUE" "CROSS"

Coral Survey
- Shape "CROSS"

Find the break
- markers 1, 2,3,4,5,6

Pinger
- active pinger 1 , 2 ,3 (1 is closest to red, 3 is closest to green, 2 in middle)

ironmig:
@maverick7170 here's the current version of the 'magic strings' file which contains all the params set / used by the missions for integrating into the GUI. https://github.com/ironmig/Navigator/blob/2d7f104e9af1195db5ea7aaebdfb29ee5f04e0d2/mission_control/navigator_missions/navigator_singleton/mission_params.yaml

mattlangford:
Happy with this? Will merge on your go ahead

ironmig:
Roger. Clear for merge

ironmig:
Never mind just noticed a typo.

maverick7170:
@ironmig Got it! Will update GUI code and put in a PR.

mattlangford:
Let me know when to pull the trigger

ironmig:
Pull it

PR #119, opened 2016-11-04
Lake day 10-30

mattlangford:
Changes David and I made on lakeday. It appears as if the TF for the left camera isn't accurate but the right camera TF is incredibly precise. 

There are some changes to the missions in here to in order to fix bugs and what not.

After this PR we will do a dev->master merge.

### DIFF
--- a/mission_control/navigator_launch/launch/subsystems/tf.launch
+++ b/mission_control/navigator_launch/launch/subsystems/tf.launch
@@ -1,18 +1,18 @@
 <launch>
-    <node pkg="tf" type="static_transform_publisher" name="measure_point_tf" args="1.2319 0 0  0 0 0 1  /base_link /measurement 100" /> <!-- Very front center of top part -->
+    <node pkg="tf" type="static_transform_publisher" name="measure_point_tf" args="1.2319 0 0  0 0 0 1  /base_link /measurement 10" /> <!-- Very front center of top part -->
     <node pkg="tf" type="static_transform_publisher" name="ins_tf" args="-0.652272 0.136525 0.192278  0 0 0 1  /measurement /ins 10" />
     
     <!-- LIDAR -->
-    <node pkg="tf" type="static_transform_publisher" name="lidar_tf" args="-0.277622 0 0.648208 0 0 0 1  /measurement /velodyne 100" />
-    <node pkg="tf" type="static_transform_publisher" name="sick_tf" args="0.5334 -0.0254 -0.6858  0 0 0 1  /measurement /sick 100" />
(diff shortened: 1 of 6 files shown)


mission_control/navigator_missions/nav_missions/coral_survey.py
```diff
@@ -19,8 +19,9 @@
 def main(navigator):
     result = navigator.fetch_result()
 
-    middle_point = np.array([-10, -70, 0]) #yield navigator.database_query("coral_survey")
-
+    #middle_point = np.array([-10, -70, 0]) 
+    resp = yield navigator.database_query("start_gate")
```
ironmig: You should check ```resp.found  == True``` here to be safe. If the database doesn't have the start gate in it, this will throw an out of index.

mission_control/navigator_launch/launch/subsystems/tf.launch
```diff
@@ -1,18 +1,18 @@
...
     <node pkg="tf" type="static_transform_publisher" name="ins_tf" args="-0.652272 0.136525 0.192278  0 0 0 1  /measurement /ins 10" />
     
     <!-- LIDAR -->
-    <node pkg="tf" type="static_transform_publisher" name="lidar_tf" args="-0.277622 0 0.648208 0 0 0 1  /measurement /velodyne 100" />
-    <node pkg="tf" type="static_transform_publisher" name="sick_tf" args="0.5334 -0.0254 -0.6858  0 0 0 1  /measurement /sick 100" />
+    <node pkg="tf" type="static_transform_publisher" name="lidar_tf" args="-0.277622 0 0.648208 0 0 0 1  /measurement /velodyne 10" />
+    <node pkg="tf" type="static_transform_publisher" name="sick_tf" args="0.5334 -0.0254 -0.6858  0 0 0 1  /measurement /sick 10" />
 
     <!-- Cameras -->
-    <node pkg="tf" type="static_transform_publisher" name="stereo_tf" args="0.017272 -0.100076 0.19812 0.56679 -0.56679 0.42278 -0.42278 /measurement /stereo_front 100" />
```
ironmig: So we're not going to have a TF for the stero frame? Also, if the right cam tf is great, why don't we just defined the left TF in reference to the right tf. We should know this transformation pretty accurately (a precise measurement of the y direction distance between them).

So something like:
```
<node pkg="tf" type="static_transform_publisher" name="stereo_left_cam_tf" args="-0.200152 0 0  0 0 0 1 / stereo_right_cam  /stereo_left_cam 100" />
```

mattlangford: @DSsoto chose to remove stereo_tf frame and position each camera individually since fudging their TF as a group gave us subpar results (meaning measurements are not perfect, or the cameras don't sit in the same spot in the tube)

PR #120, opened 2016-11-04
LAUNCH, ALARMS, SIM

mattlangford:
> New stuff and stuff.

Don't merge this before the lake day merge since it contains all that since I'm a noob and tend not to branch.

- I redid how odom is published in sim so now we actually have ecef coordinates as well as a local enu frame. This allows us to use coordinate conversion just like for the real boat (mainly for the bounds system). I updated launch files to use all that new power we have.
- I added that ogrid node back to our repo instead of just keeping it in lqrrt
- __♪ Do you see what I see?__ Now you can - there's an rviz config! Load that into rviz and see (pretty much) everything you need. (No cameras since those behave weird).
- The `station holding` alarm wasn't working as of late. Maybe this will fix it?
- Increase Hamilton references in the code to .0041215%


mission_control/navigator_launch/launch/subsystems/lqrrt_node.launch
```diff
@@ -11,12 +12,20 @@
...
     <param name="path_topic" value="/lqrrt/path"/>
     <param name="tree_topic" value="/lqrrt/tree"/>
     <param name="goal_topic" value="/lqrrt/goal"/>
     <param name="focus_topic" value="/lqrrt/focus"/>
     <param name="effort_topic" value="/lqrrt/effort"/>
   </node>
 
+  <!-- Putting this here since lqrrt needs a real ogrid in it's life -->
+  <node pkg="navigator_msg_multiplexer" type="ogrid_arbiter.py" name="ogrid_arbiter" />
+
```
tessbianchi: I like this name!

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -24,21 +24,35 @@
...
 
 class MissionResult(object):
-    def __init__(self, success=True, response="", need_rerun=False, post_function=None):
+    '''
+        The response can be used by the mission runner to determine why a missions failed (assuming it's not a normal
+            runtime issue). Add whatever response types you want.
+        Also with this class you can tell the mission runner whether you want to rerun the mission (if it failed), and
+            give an optional function to run after the mission completes (regardless of pass/fail).
+        Finally, this provides a good way to print out mission results.
+    '''
```
tessbianchi: This isn't PEP8

navigator.rviz
```diff
@@ -0,0 +1,408 @@
+Panels:
```
tessbianchi: Very nice!

mission_control/navigator_missions/navigator_singleton/navigator.py
tessbianchi: These are supposed to be double quotes.
The indentation should match the quotes
The first line needs to be Imparative and end with a period.
There should be a space between first and second line

PR #123, opened 2016-11-05
Lidar: Friday Meeting Changes

maverick7170:
This PR has the following fixes:

1. Searches for 4 totem buoys in a line at roughly 30 meters (parameters that describe a good enough line will probably need to be adjusted!)
2. Added 3 new ROI (fake) markers for the three gates around the pinger. These will automatically move to the correct locations when the 4 totems are found.
3. ROI markers will only be returned in a service request if they have been placed somewhere on the map (moved from their location at the origin). Also added the "All" name to return all fake markers in a service request.
4. If bounds are not available at startup, the bounds will default to just looking around the boat (and technically the origin).
5. Volume classifier has some rough numbers for the docking challenge. All values should be checked on next lake day!


RustyBamboo:
Looked through. Looks good. Readme doesn't have IDENTIFY DOCK.

PR #124, opened 2016-11-06
Scan the code lidar

tessbianchi:
- Makes the object database update the time stamp of the object
- Updates some aspects of the mission planner, namely adding a YAML file and adding a base mission to the mission planner, so there is always a default behavior
- Finishes ScanTheCode (finally) using edge detection and lidar backprojection


ironmig:
Would you remind rebasing this to the latest dev? It will keep the git history clean and might resolve the build issues.

mission_control/navigator_missions/nav_missions/scan_the_code.py
```diff
@@ -1,32 +1,60 @@
...
+    if not mission.stc_correct:
+        circle = navigator.move.circle_point(look_at, 8, granularity=10)
+        for p in circle:
+            if mission.stc_correct:
+                break
+            # yield p.go()
+
+    colors = yield mission.find_colors()
+    print colors
+    if colors is None:
```
ironmig: So colors is None when the perception fails? Perhaps it should throw an exception / tell the mission planner it has failed or at least log it. 

Also, does this set any params / output for what it saw besides setting the detect deliver params? I think for the judge GUI we'll want params set for what scan the code saw (#118 will have a convenient way to do this)

mission_control/navigator_missions/nodes/mission_planner.py
```diff
@@ -1,98 +1,176 @@
...
+            parent_mission.add_child(child_mission)
+
+    def _is_in_queue(self, x):
+        with self.queue.mutex:
+            return x in self.queue.queue
+
     def new_item(self, obj):
-        self.found.append(obj.name)
+        """Callback for a new object being found."""
+        print "new item, {}".format(obj.name)
```
ironmig: It would be nice if all of these prints were ros logs. This would give us the timestamp and ensure it gets saved to a file. I imagine logging is going to be very important for the mission planner.

utils/navigator_tools/navigator_tools/object_database_helper.py
```diff
@@ -0,0 +1,63 @@
+"""Use the DBHelper class to interface with the Database without having to deal with ROS things."""
```
ironmig: Good idea. Maybe even add a main function that takes some cli args?

perception/navigator_lidar_oa/src/ConnectedComponents.h
```diff
@@ -77,6 +77,7 @@ struct objectMessage
 	std::vector<geometry_msgs::Point32> strikesFrame;
 	std::vector<uint32_t> intensityPersist;
 	std::vector<uint32_t> intensityFrame;
+	std_msgs::Header header;
```
ironmig: Looks like good changes to me, but be sure to rebase with dev which contains Ira's latest PR and may already have some of these changes. Looks like you're going to get a not so fun rebase conflict in these files. Up to you to try to actually line by line resolve it or just use Ira's version.

mission_control/navigator_missions/nodes/mission_planner.py
tessbianchi: Aye

utils/navigator_tools/navigator_tools/object_database_helper.py
tessbianchi: It's not meant to be a node, but included in other python files. 

If you would like a node that has this functionality I can write one, but I think it should be a separate file that uses this class.

perception/navigator_lidar_oa/src/ConnectedComponents.h
tessbianchi: Aye

mission_control/navigator_missions/nav_missions/scan_the_code.py
ironmig: Once #118 is merged, you can set them like this
```
yield self.navigator.mission_params["scan_the_code_color1"].set("YELLOW")
yield self.navigator.mission_params["scan_the_code_color2"].set("BLUE")
yield self.navigator.mission_params["scan_the_code_color2"].set("GREEN")
```

mission_control/navigator_missions/nodes/mission_planner.py
tessbianchi: txros does not have logging.

I will add a logging debugging tool in the next PR.

tessbianchi:
Read to be merged

ironmig:
Looks like semaphore still isn't happy. 
```
  File "<string>", line 8, in <module>
  File "/home/runner/mil_ws/src/Navigator/utils/navigator_tools/navigator_tools/object_database_helper.py", line 2, in <module>
    from navigator_msgs.msg import PerceptionObjects
ImportError: cannot import name PerceptionObjects
```
Think it should be ```PerceptionObjectArray```

mission_control/navigator_missions/nav_missions/scan_the_code.py
tessbianchi: Will do

tessbianchi:
Yes, it's fixed

PR #128, opened 2016-11-11
 SIM: Sandbox Simulator Part 2

whispercoros:
This PR is the second and final iteration of the sandbox world. It adds functional versions of the following challenges:
    Scan the Code (adds three faces on a platform model)
    Identify Symbols and Dock (fixes lighting issues and unifies dock model)
    Coral Survey
    Find the Break

I have also taken care of a few housekeeping tasks, such as cleaning up the file structure, removing the unnecessary ray projections from the Sick LiDAR, and adding a down camera to the model of the boat. The simulated down camera's location is roughly the same as where the mechanical team is planning to mount the physical one.


mattlangford:
Everything looks good here

tessbianchi:
Yay!

PR #138, opened 2016-11-15
REMOTE_CONTROL: Controller library and keyboard control

whispercoros:
This PR mainly cleaned up and refactored the remote control system. With the creation of the controller library, any basic interface to a controller (GUI, keyboard, joystick, etc.) can access all of our remote control functions without having to re-implement them locally.

Keyboard control was implemented using a client-server architecture in which one user at a time can 'lock' control of the server by pressing Shift+L. The client does not unlock control, control can be taken by a new keyboard at any time (this is to avoid someone locking control and then leaving without unlocking). All of the key bindings are printed in a help menu. It can be run in a terminal with this command (this includes the terminal emulator in RQT so that the user can drive with cameras or RVIZ in the same window):

    rosrun navigator_keyboard_control navigator_keyboard_client.py

A minor install script patch as well as the wrench_arbiter's /wrench/current node (from lake day 11/13/16) are also included.

### DIFF
--- a/utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
+++ b/utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
@@ -1,62 +1,39 @@
 #!/usr/bin/env python
 
 
-import itertools
-
-from geometry_msgs.msg import WrenchStamped
-from kill_handling.broadcaster import KillBroadcaster
-from nav_msgs.msg import Odometry
-from navigator_alarm import AlarmBroadcaster
-from navigator_msgs.srv import WrenchSelect, ShooterManual, ShooterManualRequest
-from navigator_msgs.msg import ShooterDoAction, ShooterDoActionGoal
-import rospy
 import actionlib
+from navigator_msgs.msg import ShooterDoAction, ShooterDoActionGoal
+from remote_control_lib import RemoteControl
+import rospy
 from sensor_msgs.msg import Joy
-from std_srvs.srv import SetBool, SetBoolRequest, Trigger, TriggerRequest
+from std_srvs.srv import Trigger, TriggerRequest
 
-rospy.init_node("joystick")
+
+__author__ = "Anthony Olive"
+__maintainer__ = "Anthony Olive"
+__email__ = "anthony@iris-systems.net"
+__copyright__ = "Copyright 2016, MIL"
+__license__ = "MIT"
+
+
+rospy.init_node('joystick')
 
 
 class Joystick(object):
 
     def __init__(self):
-
         self.force_scale = rospy.get_param("~force_scale", 600)
         self.torque_scale = rospy.get_param("~torque_scale", 500)
 
-        self.wrench_choices = itertools.cycle(['rc', 'autonomous'])
-        self.current_pose = Odometry()
-
-        self.active = False
-
-        self.alarm_broadcaster = AlarmBroadcaster()
-        self.kill_alarm = self.alarm_broadcaster.add_alarm(
-            name='kill',
-            action_required=True,
-            severity=0
-        )
-
-        self.station_hold = self.alarm_broadcaster.add_alarm(
-            name='station_hold',
-            action_required=False,
-            severity=3
-        )
-
-        # self.docking_alarm = self.alarm_broadcaster.add_alarm(
-        #     name='docking',
-        #     action_required=True,
-        #     severity=0
-        # )
-
-        self.wrench_pub = rospy.Publisher("/wrench/rc", WrenchStamped, queue_size=1)
-        # rospy.wait_for_service('/change_wrench')
-        self.wrench_changer = rospy.ServiceProxy('/change_wrench', WrenchSelect)
+        self.remote = RemoteControl('joystick', "/wrench/rc")
 
         self.shooter_fire_client = actionlib.SimpleActionClient('/shooter/fire', ShooterDoAction)
         self.shooter_load_client = actionlib.SimpleActionClient('/shooter/load', ShooterDoAction)
         self.shooter_cancel_client = rospy.ServiceProxy('/shooter/cancel', Trigger)
 
-        rospy.Subscriber("joy", Joy, self.joy_cb)
+        rospy.Subscriber("joy", Joy, self.joy_recieved)
+
+        self.active = False
         self.reset()
 
     def reset(self):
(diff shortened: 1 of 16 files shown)


utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_server.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class KeyboardServer(object):
+
+    def __init__(self):
+        self.force_scale = rospy.get_param("/joystick_wrench/force_scale", 600)
+        self.torque_scale = rospy.get_param("/joystick_wrench/torque_scale", 500)
+
+        self.remote = RemoteControl('keyboard', "/wrench/keyboard")
+        rospy.Service('/keyboard_control', KeyboardControl, self.key_recieved)
```
mattlangford: It seems like services aren't the way to do this. What motivated you to use a service instead of a topic

utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_server.py
```diff
@@ -0,0 +1,116 @@
...
+        # If the key was from the client with locked control, pass it to execution
+        elif (req.uuid == self.locked_uuid):
+            self.execute_key(req.keycode)
+            return {"generated_uuid": '', "is_locked": True}
+
+        # Ignore keys sent by a client that has not locked control of the service
+        else:
+            return {"generated_uuid": '', "is_locked": False}
+
+    def execute_key(self, key):
```
mattlangford: As I always say, there's almost always a "cleaner" way to do things rather than using a lot of `elif`'s.
For example, you could have a dictionary that maps key's to functions (`self.handle_key = {ord('k'): self.remote.toggle_kill, ...}`). That way this function would be one line: `self.handle_key[key]()`.

Not asking for a change just suggesting

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
```diff
@@ -1,62 +1,31 @@
...
-from navigator_msgs.msg import ShooterDoAction, ShooterDoActionGoal
+from remote_control_lib import RemoteControl
 import rospy
-import actionlib
 from sensor_msgs.msg import Joy
-from std_srvs.srv import SetBool, SetBoolRequest, Trigger, TriggerRequest
 
-rospy.init_node("joystick")
+
+__author__ = "Anthony Olive"
```
whispercoros: I've rewritten nearly every line in this file by now, but I'll remove it if you like. It's mainly there so that everyone knows who to yell at when things break.

utils/remote_control/navigator_keyboard_control/remote_control_lib/remote_control_lib.py
```diff
@@ -0,0 +1,219 @@
...
+
+        self.shooter_load_client = actionlib.SimpleActionClient('/shooter/load', ShooterDoAction)
+        self.shooter_fire_client = actionlib.SimpleActionClient('/shooter/fire', ShooterDoAction)
+        self.shooter_cancel_client = rospy.ServiceProxy('/shooter/cancel', Trigger)
+
+        self.is_killed = False
+        self.is_timed_out = False
+        self.clear_wrench()
+
+    def timeout_check(function):
```
whispercoros: Since this is fairly controller specific, it's the controller's job to let the library know that it is timed out. Once this happens, most functions are effectively disabled. If you would like to see an example of this, check out  utils/navigator_gui/navigator_gui/dashboard.py ~line 335.

utils/remote_control/navigator_keyboard_control/remote_control_lib/remote_control_lib.py
```diff
@@ -0,0 +1,219 @@
...
+            return function(self, *args, **kw)
+        return wrapper
+
+    def update_kill_status(self, alarm):
+        '''
+        Updates the kill status display when there is an update on the kill
+        alarm. Caches the last displayed kill status to avoid updating the
+        display with the same information twice.
+        '''
+        if (alarm.clear):
```
whispercoros: Good point, will change this.

utils/remote_control/navigator_keyboard_control/remote_control_lib/remote_control_lib.py
```diff
@@ -0,0 +1,219 @@
...
+            self.is_killed = False
+        else:
+            self.is_killed = True
+
+    @timeout_check
+    def kill(self, *args, **kw):
```
whispercoros: Will change this.

utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_client.py
```diff
@@ -0,0 +1,156 @@
...
+    def read_key(self):
+        '''
+        Reads the newest key from the buffer and discards all old keys that
+        were not read.
+        '''
+        keycode = -1
+        new_keycode = self.screen.getch()
+
+        # This eliminates building a buffer of keys that takes forever to process
+        while(new_keycode != -1):
```
whispercoros: This while loop will only be cycling through a very small amount of keys (at most 10 for most keyboards), and will only be run when it is called by the spinner. I agree that checking for shutdown is good convention though, so I will add it.

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
tessbianchi: yeah?
yeah? 
yeah?
yeah?

utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_server.py
whispercoros: I cannot see an instance where a topic would be better? I am assuming of course that you mean a topic that sits on the client side and publishes key presses. The issue with this is only one keyboard client would be able to publish to the topic. In this instance, would any running client not be killed when another is spooled up? By using a service, multiple keyboard clients can be running on multiple machines and the locked client can be easily switched at any time (i.e. easy and quick transfer of control).

utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_server.py
whispercoros: I will likely make this change along with the others as I am a sucker for efficient code.

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
whispercoros: This line was not intended to be the one you yell at me about, though!

utils/remote_control/navigator_keyboard_control/nodes/navigator_keyboard_server.py
whispercoros: I think you misunderstood the purpose of the file. Only one instance of the keyboard server is run and it is run on navigator. The keyboard clients operate as service clients and do not publish to anything or offer any service. This seems to be exactly what you described here:

    "...multiple service clients attached to the same server."

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
whispercoros: @DSsoto and @jnez71 sided with me on this, leaving it as is

PR #140, opened 2016-11-15
OGRID, ALARMS, MISSION

mattlangford:
Lots of changes to important stuff.

- The ogrid fusing node and bounds server have been converted to dynamic reconfig which will let us change the values really easily in HI.
- There's a colorizer for buoys in the database. It's still in development and this week we should have a much better version to test.
- Start gate seems to work in sim with pairs of gates, not so much with only two.
- Lidar classification changes as of the last lake day

__Note to those publishing ogrids__:
- They need to be square for now, there's a bug that causes weird issues that get worse the more the height and width differ.
- The resolution still needs to be set to 0.3, scaling isn't really high on my priority list of things to fix right now

### DIFF
--- a/mission_control/navigator_missions/nav_missions/coral_survey.py
+++ b/mission_control/navigator_missions/nav_missions/coral_survey.py
@@ -6,7 +6,7 @@ import tf
 import tf.transformations as trns
 import numpy as np
 import navigator_tools
-from navigator_tools import fprint, MissingPerceptionObject
+from navigator_tools import fprint
 from navigator_singleton import pose_editor
 from twisted.internet import defer
 
@@ -21,22 +21,19 @@ def main(navigator):
 
     #middle_point = np.array([-10, -70, 0]) 
     est_coral_survey = yield navigator.database_query("Coral_Survey")
-    if not est_coral_survey.found:
-        raise MissingPerceptionObject("Coral_Survey")
+    assert est_coral_survey.found, "Coral Survey ROI not found."
     
     yield navigator.move.set_position(est_coral_survey.objects[0]).go()
 
     totem = yield navigator.database_query("totem")
-    if not totem.found:
-        raise MissingPerceptionObject("totem")
+    assert totem.found, "No totem objects"
     
     # Get the closest totem object to the boat
     totem_np = map(lambda obj: navigator_tools.point_to_numpy(obj), totem.objects)
     dist = map(lambda totem_np: np.linalg.norm(totem_np - navigator_tools.point_to_numpy(est_coral_survey.objects[0])), totems_np)
     
     # Complain if the buoy is too far away
-    if not min(dist) > 20:  # m
-        raise MissingPerceptionObject("totem")
+    assert min(dist) < 20, "Totems are too far away from Coral Survey"
     
     middle_point = navigator_tools.point_to_numpy(totem.objects[0].position)
     quads_to_search = [1, 2, 3, 4]
--- a/mission_control/navigator_missions/nav_missions/coral_survey.py
+++ b/mission_control/navigator_missions/nav_missions/coral_survey.py
@@ -6,7 +6,7 @@ import tf
 import tf.transformations as trns
 import numpy as np
 import navigator_tools
-from navigator_tools import fprint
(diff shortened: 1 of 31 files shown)


mission_control/navigator_missions/nav_missions/start_gate.py
```diff
@@ -172,6 +174,8 @@ def make_ogrid_transform(self):
                                                               [0, 0, 0, 1])
         dx = self.x_upper - self.x_lower
         dy = self.y_upper - self.y_lower
+        _max = max(dx, dy)
+        dx = dy = _max
```
mattlangford: Hopefully this will be fixed. I need to get some time to fix the ogrid fusing script

perception/navigator_vision/nodes/database_color.py
```diff
@@ -0,0 +1,351 @@
...
+        self.hue_error_reject = .4  # rads
+
+        # To decide when to accept color observations
+        self.error_tolerance = .4 # rads
+        self.spottings_req = 4
+        self.similarity_reject = .1 # If two colors are within this amount of error, reject
+
+        rospy.Timer(ros_t(self.update_time), self.do_update)
+
+    def valid_color(self, _id):
```
mattlangford: The logic here is going to change soon...

whispercoros:
@jnez71

mission_control/navigator_missions/nav_missions/coral_survey.py
```diff
@@ -20,14 +20,22 @@ def main(navigator):
...
-    resp = yield navigator.database_query("coral_survey")
-    if not resp.found:
-        result.success = False
-        result.response = result.DbObjectNotFound
-        result.message = "Coral survey not found in the database."
-        defer.returnValue(result)
-
-    middle_point = navigator_tools.point_to_numpy(resp.objects[0].position)
+    est_coral_survey = yield navigator.database_query("Coral_Survey")
+    assert est_coral_survey.found, "Coral Survey ROI not found."
```
tessbianchi: There is no need for this now.... Maybe change in the next PR. This will cause the mission to fail in the future.

mission_control/navigator_missions/nav_missions/coral_survey.py
```diff
@@ -20,14 +20,22 @@ def main(navigator):
...
+    
+    yield navigator.move.set_position(est_coral_survey.objects[0]).go()
+
+    totem = yield navigator.database_query("totem")
+    assert totem.found, "No totem objects"
```
tessbianchi: Same here.

mission_control/navigator_missions/nav_missions/coral_survey.py
```diff
@@ -20,14 +20,22 @@ def main(navigator):
...
+    
+    # Get the closest totem object to the boat
+    totem_np = map(lambda obj: navigator_tools.point_to_numpy(obj), totem.objects)
+    dist = map(lambda totem_np: np.linalg.norm(totem_np - navigator_tools.point_to_numpy(est_coral_survey.objects[0])), totems_np)
+    
+    # Complain if the buoy is too far away
+    assert min(dist) < 20, "Totems are too far away from Coral Survey"
+
```
tessbianchi: Since this is a mission, we shouldn't really error handle this way. 

This is the top of the stack and errors need to be handled here, not thrown, or the mission will not be repeated. (Unless its a MissingPerceptionObject message)
This is the top of the stack and errors need to be handled here, not thrown. (Unless its a MissingPerceptionObject message)

perception/navigator_vision/nodes/database_color.py
```diff
@@ -0,0 +1,351 @@
+#!/usr/bin/env python
```
tessbianchi: This is great!

It would be nice if you moved the ImageHolder and Debug to navigator_tools so others could use them.

But not necessary!

mission_control/navigator_missions/nav_missions/coral_survey.py
mattlangford: Agreed

mattlangford:
Good to merge?

tessbianchi:
Yes. Merging.

PR #142, opened 2016-11-17
COMMAND: Circles!

mattlangford:
Adds support for lqrrt circles. Also fixes most of the gazebo issues I've been having.

__Some important things to note for missions__
- The move formally known as `circle_point` no longer circles a point in discrete move commands like it used to. To keep the same functionality as before (discrete move commands) use `d_circle_point`. The new `circle_point` continuously circles a point at the boat's current distance to the point and the relative orientation to the point. This move will not actively avoid obstacles and will instead return a failure if an obstacle is detected.
- `navigator.move.(...).go()` now returns the action result. The action result has a `failure_reason` attribute which is normally blank (unless something went wrong). 
- Let me know if you have any other questions!

### DIFF
--- a/mission_control/navigator_missions/navigator_singleton/pose_editor.py
+++ b/mission_control/navigator_missions/navigator_singleton/pose_editor.py
@@ -91,27 +91,55 @@ def get_valid_point(nav, point):
         # TODO: Make boat go to edge
         return nav.pose[0]
 
+
+class Spiraler(object):
+    def __init__(self, pe, point, *args, **kwargs):
+        self.pe = pe
+        self.point = point
+
+        self.mpr = kwargs.get("mpr", 0)
+        self.direction = kwargs.get("direction", "cw")
+        self.revolutions = kwargs.get("revolutions", 0)
+
+
 class PoseEditor2(object):
-    '''
+    """
     Used to chain movements together
+
     ex:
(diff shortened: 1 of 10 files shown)


mattlangford:
I'd like this not to get merged until we do a dev->master merge

mission_control/navigator_missions/navigator_singleton/pose_editor.py
```diff
@@ -216,16 +230,41 @@ def to_lat_long(self, lat, lon, alt=0):
...
+        """
+        Sprials an ENU point.
+
+        `point` is a 2d or 3d numpy point or geometery_msgs/PointStamped to spiral
+        `direction` can be 'cw' or 'ccw' for clockwise or counter clockwise
+        `meters_per_rev` is the number of meters per revolution to increase the radius of the spiral
+        `revolutions` is the number of revolutions to complete
+
+        NOTE: Don't use this function with other pose editor functions like you traditionally would
+        """
```
tessbianchi: So do not chain them? Is this what you are saying?

mattlangford: Yeah, don't do `navigator.move.yaw_left(...).circle_point(...).forward(...).go()`
Circling and spiraling is handled differently at the planner level and so it should be treated differently here too.

PR #145, opened 2016-11-18
DETECT DELIVER: Lakeday fixes, better alignment

ironmig:
- start the flywheels spinning up when the shooter starts loading
- add TF for shooter (where the ball leaves the flywheels), so we can align based on shooter location not base_link
- Clean camera_to_lidar so it does not bring great shame to my family
- Matthew langford fixed a bug with importing something something __init
- Can set a param to align to the big or small goal in detect deliver

### DIFF
--- a/mission_control/navigator_missions/nav_missions/detect_deliver.py
+++ b/mission_control/navigator_missions/nav_missions/detect_deliver.py
@@ -55,13 +55,13 @@ class DetectDeliverMission:
     def get_waypoint(self):
         res = yield self.navigator.database_query("shooter")
         if not res.found:
-            print "Waypoint not found in database, exiting"
+            fprint("shooter waypoint not found", title="DETECT DELIVER",  msg_color='red')
             raise Exception('Waypoint not found')
         self.waypoint_res = res
 
     @txros.util.cancellableInlineCallbacks
     def circle_search(self):
-        fprint("Starting Circle Search", title="DETECT DELIVER",  msg_color='yellow')
+        fprint("Starting Circle Search", title="DETECT DELIVER",  msg_color='green')
         #yield self.navigator.move.look_at(navigator_tools.rosmsg_to_numpy(self.waypoint_res.objects[0].position)).go()
         pattern = self.navigator.move.circle_point(navigator_tools.rosmsg_to_numpy(
             self.waypoint_res.objects[0].position), radius=self.circle_radius,  theta_offset=self.theta_offset)
@@ -69,15 +69,16 @@ class DetectDeliverMission:
         searcher = self.navigator.search(
             vision_proxy='get_shape', search_pattern=pattern, Shape=self.Shape, Color=self.Color)
         yield searcher.start_search(timeout=self.search_timeout_seconds, spotings_req=self.spotings_req, move_type="skid")
-        fprint("Ended Circle Search", title="DETECT DELIVER",  msg_color='yellow')
+        fprint("Ended Circle Search", title="DETECT DELIVER",  msg_color='green')
 
     @txros.util.cancellableInlineCallbacks
     def align_to_target(self):
+        print "fff"
         self.markers = MarkerArray()
         while not (yield self.is_found()):
-          fprint("Shape not found Error={}".format(self.resp.error), title="DETECT DELIVER", msg_color='yellow')
+            fprint("Shape not found Error={}".format(self.resp.error), title="DETECT DELIVER", msg_color='red')
         while not (yield self.get_normal()):
-          fprint("Normal found Error={}".format(self.normal_res.error), title="DETECT DELIVER", msg_color='yellow')
+            fprint("Normal found Error={}".format(self.normal_res.error), title="DETECT DELIVER", msg_color='red')
         yield self.get_aligned_pos()
 
         move = self.navigator.move.set_position(self.aligned_position).set_orientation(self.aligned_orientation).forward(self.target_offset_meters)
@@ -98,7 +99,7 @@ class DetectDeliverMission:
         move_marker.color.a = 1
         self.markers.markers.append(move_marker)
         yield self.markers_pub.publish(self.markers)
-        fprint("Aligning to shoot at {}".format(move), title="DETECT DELIVER", msg_color='yellow')
+        fprint("Aligning to shoot at {}".format(move), title="DETECT DELIVER", msg_color='red')
         yield move.go(move_type="skid")
 
 
--- a/mission_control/navigator_missions/nav_missions/detect_deliver.py
+++ b/mission_control/navigator_missions/nav_missions/detect_deliver.py
@@ -13,7 +13,7 @@ from twisted.internet import defer
 from image_geometry import PinholeCameraModel
 from visualization_msgs.msg import Marker,MarkerArray
 import navigator_tools
-from navigator_tools import fprint
+from navigator_tools import fprint, MissingPerceptionObject
 
 class DetectDeliverMission:
     # Note, this will be changed when the shooter switches to actionlib
@@ -32,7 +32,6 @@ class DetectDeliverMission:
             self.navigator.nh, '/shooter/load', ShooterDoAction)
         self.shooterFire = txros.action.ActionClient(
             self.navigator.nh, '/shooter/fire', ShooterDoAction)
-        self.markers_pub = self.navigator.nh.advertise("/detect_deliver/debug_marker",MarkerArray)
 
     def _bounding_rect(self,points):
         np_points = map(navigator_tools.point_to_numpy, points)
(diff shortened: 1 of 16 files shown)


mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -42,95 +42,78 @@ def _bounding_rect(self,points):
...
-        marker.text = "Offset goal"
-        marker.ns = "detect_deliver"
-        marker.id = 3
-        marker.color.b = 1
-        marker.color.a = 1
-        self.markers.markers.append(marker)
-        yield move
+        move = self.navigator.move.set_position(self.aligned_position).set_orientation(self.aligned_orientation).forward(self.target_offset_meters)
+        move = move.left(-self.shooter_baselink_tf._p[1]).forward(-self.shooter_baselink_tf._p[0]) #Adjust for location of shooter
+        move_marker = Marker()
```
mattlangford: Marker is redundant

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -42,95 +42,78 @@ def _bounding_rect(self,points):
...
     def get_waypoint(self):
         res = yield self.navigator.database_query("shooter")
         if not res.found:
-            print "Waypoint not found in database, exiting"
+            fprint("shooter waypoint not found", title="DETECT DELIVER",  msg_color='red')
             raise Exception('Waypoint not found')
         self.waypoint_res = res
 
     @txros.util.cancellableInlineCallbacks
     def circle_search(self):
```
mattlangford: This function will not work anymore, for the same functionality, use `...move.d_circle_point(...)` instead

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -42,95 +42,78 @@ def _bounding_rect(self,points):
...
         yield next(pattern).go()
         searcher = self.navigator.search(
             vision_proxy='get_shape', search_pattern=pattern, Shape=self.Shape, Color=self.Color)
         yield searcher.start_search(timeout=self.search_timeout_seconds, spotings_req=self.spotings_req, move_type="skid")
-        print "Ended Circle Search"
+        fprint("Ended Circle Search", title="DETECT DELIVER",  msg_color='green')
 
     @txros.util.cancellableInlineCallbacks
     def align_to_target(self):
+        print "fff"
```
mattlangford: fprint some info here instead of f's

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -42,95 +42,78 @@ def _bounding_rect(self,points):
...
         self.markers = MarkerArray()
         while not (yield self.is_found()):
-          print "Shape Not Found, Error: ", self.resp.error
+            fprint("Shape not found Error={}".format(self.resp.error), title="DETECT DELIVER", msg_color='red')
```
mattlangford: Bingo, bango

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -181,13 +165,21 @@ def find_and_shoot(self):
...
         yield self.circle_search()  # Go to waypoint and circle until target found
         yield self.align_to_target()
-        yield self.offset_for_target()  # Move a little bit forward to be centered to target
         yield self.shoot_all_balls()
         yield self.navigator.vision_proxies["get_shape"].stop()
 
 @txros.util.cancellableInlineCallbacks
+def setup_mission(navigator):
+    color = yield navigator.mission_params["scan_the_code_color3"].get()
+    fprint("Setting search color to scan the code color 3 = {}".format(color), title="DETECT DELIVER",  msg_color='green')
```
mattlangford: This isn't really a critical thing but I feel like it'd be good to enforce some standards with what to print in which color with fprint. Like green should be for things that could fail but succeed instead. I'd imagine these info prints would be blue? What are your thoughts

mission_control/navigator_missions/navigator_singleton/__init__.py
```diff
@@ -2,9 +2,9 @@
 import os
 
 for module in os.listdir(os.path.dirname(__file__)):
-    if module == '__init__.py' or module[-3:] != '.py' and module[-1] != '~':
+    if module == '__init__.py' or(module[-3:] != '.py' or module[-1] == '~':
```
mattlangford: Does it work with this extra parathenisis here?

mission_systems/shooter/nodes/ShooterControl.py
```diff
@@ -41,9 +42,28 @@ def __init__(self):
         self.status = "Standby"
         self.status_pub = rospy.Publisher('/shooter/status', String, queue_size=5)
         self.manual_used = False
+        self.killed = False
+        self.kill_listener = AlarmListener("kill", self.update_kill_status)
+
```
mattlangford: Nice

ironmig:
Comments addressed, remained using granular search but added commented out new circle mode.

ironmig:
Updated with all my stuff from lakeday and some bag-tested changes. Look me over!

mission_control/navigator_missions/nav_missions/pinger.py
```diff
@@ -49,10 +54,14 @@ def get_objects(self):
...
 
+    @txros.util.cancellableInlineCallbacks
     def get_observation_poses(self):
         """Set 2 points to observe the pinger from, in front of gates 1 and 3"""
         self.get_gate_perp()
-        pose = yield self.navigator.tx_pose[:2]
+        #Make sure they are actually in a line
+        if np.isnan(self.g_perp[0]) or np.isnan(self.g_perp[1]):
+            raise Exception("Gates are not in a line")
+        pose = (yield self.navigator.tx_pose)[0][:2]
```
mattlangford: If this is the only thing keeping this function from being synchronous? There's a secret `navigator.pose` that gets updated on each odom message so you can use that instead of decorating this function.

I feel like the less defer functions the cleaner the code.

mission_control/navigator_missions/nav_missions/pinger.py
```diff
@@ -68,6 +77,7 @@ def get_observation_poses(self):
     @txros.util.cancellableInlineCallbacks
     def search_samples(self):
         """Move to each observation point and listen to the pinger while sitting still"""
+        yield self.get_observation_poses()
```
mattlangford: (Assuming `self.observation_points` and `self.look_at_points` are only used here) Any reason not to do:
```python
obs, looks = self.get_observation_poses()  # returns [obs_points, look_at_points]
for obs, look in zip(obs, looks):
    ...
    ...
```

Setting acouple class variables here doesn't make sense if you're only using it in this one place right?

mission_control/navigator_missions/nav_missions/pinger.py
```diff
@@ -91,18 +101,116 @@ def go_thru_gate(self):
...
+                color_1 = "RED"
+                yield self.new_marker(position=navigator_tools.rosmsg_to_numpy(sorted_1[0].position), color=(1,0,0))
+            elif sorted_1[0].color.g > 0.9:
+                color_1 = "GREEN"
+                yield self.new_marker(position=navigator_tools.rosmsg_to_numpy(sorted_1[0].position), color=(0,1,0))
+            else:
+                color_1 = "UNKNOWN"
+                yield self.new_marker(position=navigator_tools.rosmsg_to_numpy(sorted_1[0].position), color=(1,1,1))
+
+            if int(self.gate_index) == 0:
```
mattlangford: Could `gate_index` not be an int?

mattlangford:
great work Kevin!

mission_control/navigator_missions/nav_missions/pinger.py
ironmig: It's a numpy int thing. I'm just dotting my i's

utils/navigator_tools/nodes/video_player.py
```diff
@@ -36,17 +36,33 @@ def __init__(self):
...
+
       self.paused = False
       self.ended = False
       self.last_key = 255
-      
+
       if (self.slider):
           cv2.namedWindow('Video Control')
-          cv2.createTrackbar('Frame','Video Control',int(0),int(self.num_frames),self.trackbar_cb)
+          cv2.createTrackbar('Frame','Video Control',int(self.start_frame),int(self.num_frames),lambda d: self.cap.set(cv2.cv.CV_CAP_PROP_POS_FRAMES,d))
```
mattlangford: You can't set variables like this in a lambda.
What I would suggest:
```python
self_set = lambda var, value: setattrt("self", var, value)
# then this goes in the callback
lambda d: selfset("roi_x_offset", d)
```

ironmig:
Fixed!

PR #146, opened 2016-11-18
REMOTE and GUI: Creates a GUI to control the shooter manually

whispercoros:
This fixes the issues brought up iin #144 for use with the GUI. The remote control library can now execute all of the automatic control commands for the shooter mechanism (load, fire, and cancel) and a few of the manual ones as well (linear actuator extension and retraction, setting the speed of the acceleration discs, and reseting the automatic control service). More important than any of the features previously mentioned, this PR also greatly increases the repositories Hamiltonian coefficient.


utils/navigator_gui/navigator_gui/shooter.py
```diff
@@ -0,0 +1,148 @@
...
+        self.remote.set_disc_speed(self.disc_speed_setting)
+
+    def update_kill_status(self, alarm):
+        '''
+        Updates the kill status display when there is an update on the kill
+        alarm.
+        '''
+        self.is_killed = not alarm.clear
+
+    def update_shooter_status(self, msg):
```
ironmig: Do you want this published at a fixed interval or whenever something changes? Also, please add a yellow "manual"  option to be set when manual control is used but is has not yet been canceled or reset.

utils/navigator_gui/navigator_gui/shooter.py
```diff
@@ -0,0 +1,148 @@
...
+
+        self.shooter_status = {
+            "received": "Unknown",
+            "cached": "Unknown"
+        }
+        self.disc_speed_setting = 0
+
+        self.connect_ui()
+
+        self.kill_listener = AlarmListener("kill", self.update_kill_status)
```
ironmig: Do you use this kill listener for the shooter? For the remote control lib, might be a good idea to not allow shooter control when killed / auto cancel.

utils/navigator_gui/navigator_gui/dashboard.py
```diff
@@ -156,8 +154,7 @@ def connect_ros(self):
         rospy.Subscriber("/battery_monitor", Float32, self.cache_battery_voltage)
         rospy.Subscriber("/clock", Clock, self.cache_system_time)
 
-        self.wrench_changer = rospy.ServiceProxy('/change_wrench', WrenchSelect)
```
whispercoros: After the actual wrench changing was moved to the remote control library and the service call method of checking it's status was switched to caching messages from a publisher, this was an unused service client.

utils/navigator_gui/navigator_gui/shooter.py
whispercoros: This is not implemented in the remote control library. None of the functions there are dependent on the killed state because they assume the user is not trying to do something stupid (e.g. there may be a scenario where we want to be able to test the shooter while killed). If you think this is a good behavior to implement, I think it would be best to leave it on the actual services that run on Navigator.

whispercoros: I should also point out that upon reading this, I realized that there was a callback to monitor the killed status in the shooter GUI. I have removed it since it was not used anywhere.

ironmig:
```
diff --git a/mission_systems/shooter/nodes/ShooterControl.py b/mission_systems/shooter/nodes/ShooterControl.py
index ab9a488..69d5bf6 100755
--- a/mission_systems/shooter/nodes/ShooterControl.py
+++ b/mission_systems/shooter/nodes/ShooterControl.py
@@ -7,6 +7,7 @@ import time
 from navigator_msgs.msg import ShooterDoAction, ShooterDoActionFeedback, ShooterDoActionResult, ShooterDoActionGoal
 from navigator_msgs.srv import ShooterManual, ShooterManualResponse
 from std_srvs.srv import Trigger,TriggerResponse
+from std_msgs.msg import String
 
 class ShooterControl:
     def __init__(self):
@@ -25,6 +26,7 @@ class ShooterControl:
         self.load_total_time = self.load_retract_time + self.load_pause_time + self.load_extend_time
         self.fire_retract_time = rospy.Duration(0, rospy.get_param('~controller/fire/retract_time_millis', 400)*1000000)
         self.fire_shoot_time = rospy.Duration(0, rospy.get_param('~controller/fire/shoot_time_millis', 1000)*1000000)
+        
         self.total_fire_time = max(self.fire_shoot_time,self.fire_retract_time)
         self.loaded = False #Assume linear actuator starts forward
         self.stop = False
@@ -37,9 +39,10 @@ class ShooterControl:
         self.cancel_service = rospy.Service('/shooter/cancel', Trigger, self.cancel_callback)
         self.manual_service = rospy.Service('/shooter/manual', ShooterManual, self.manual_callback)
         self.reset_service = rospy.Service('/shooter/reset', Trigger, self.reset_callback)
+        self.status = "Standby"
+        self.status_pub = rospy.Publisher('/shooter/status', String, queue_size=5)
         self.manual_used = False
 
-
     def load_execute_callback(self, goal):
         result = ShooterDoActionResult()
         if self.manual_used:
@@ -66,6 +69,7 @@ class ShooterControl:
         rate = rospy.Rate(50) # 50hz
         feedback = ShooterDoActionFeedback()
         self.motor_controller.setMotor2(-1)
+        self.status = "Loading"
         while dur_from_start < self.load_total_time and not self.stop:
             dur_from_start = rospy.get_rostime() - start_time
             feedback.feedback.time_remaining = self.load_total_time - dur_from_start
@@ -90,6 +94,7 @@ class ShooterControl:
             self.stop = False
             self.load_server.set_preempted (result=result.result)
             return
+        self.status = "Loaded"
         self.motor_controller.setMotor1(0)
         result.result.success = True
         self.loaded = True
@@ -124,6 +129,7 @@ class ShooterControl:
         feedback = ShooterDoActionFeedback()
         self.motor_controller.setMotor1(1.0)
         self.motor_controller.setMotor2(-1.0)
+        self.status = "Firing"
         while dur_from_start < self.total_fire_time and not self.stop:
             dur_from_start = rospy.get_rostime() - start_time
             feedback.feedback.time_remaining = self.total_fire_time - dur_from_start
@@ -140,6 +146,7 @@ class ShooterControl:
             self.stop = False
             self.fire_server.set_preempted (result=result.result)
             return
+        self.status = "Standby"
         self.motor_controller.setMotor1(0)
         self.motor_controller.setMotor2(0)
         result.result.success = True
@@ -158,6 +165,7 @@ class ShooterControl:
           self.motor2_stop = 0
 
     def cancel_callback(self, req):
+        self.status = "Canceled"
         self.manual_used = True
         self.motor1_stop = 0
         self.motor2_stop = 0
@@ -166,6 +174,7 @@ class ShooterControl:
         return TriggerResponse(success=True)
 
     def manual_callback(self, req):
+        self.status = "Manual"
         self.manual_used = True
         self.motor1_stop = -req.feeder
         self.motor2_stop = req.shooter
@@ -175,13 +184,20 @@ class ShooterControl:
         return res
 
     def reset_callback(self,data):
+        self.status = "Standby"
         self.stop_actions()
         self.manual_used = False
+        self.loaded = False
         return TriggerResponse(success=True)
 
+    def run(self):
+        r = rospy.Rate(5) # 10hz
+        while not rospy.is_shutdown():
+            self.status_pub.publish(self.status)
+            r.sleep()
+
 
 if __name__ == '__main__':
     rospy.init_node('shooter_control')
     control = ShooterControl()
-    #rate = rospy.Rate(10) # 10hz
-    rospy.spin()
+    control.run()
```

utils/navigator_gui/navigator_gui/shooter.py
whispercoros: It really doesn't effect the GUI operation either way. Manual mode has been added, but all statuses are assumed to be capitalized (I just noticed yours was lowercase).

utils/navigator_gui/navigator_gui/shooter.py
mattlangford: It'd be good to have the shooter not work while the boat is killed. Seems like that'd be something to put into the shooter code though.

PR #147, opened 2016-11-18
BAG: Add fake action server

mattlangford:
Just like the title implies, this adds a fake action server to the bag launch that lets you run missions on bags and see where the boat wanted to go and then return that it got there after some period of time.
Bag launch flags:
- `oa`:  default: `True`, used to start up lidar database and ogrid arbiter
- `im_proc_sr`: default : `False`, used to start up image proc for stereo_right camera
- `im_proc_sl`: default : `False`, used to start up image proc for stereo_left camera
- `im_proc_rr`: default : `False`, used to start up image proc for right camera


mission_control/navigator_launch/launch/bag.launch
```diff
@@ -1,10 +1,16 @@
 <launch>
```
ironmig: Some more things we might want in here to test with bags real easy

- optionally launch perception.launch
- optionally launch shooter

utils/navigator_tools/navigator_tools/general_helpers.py
```diff
@@ -8,17 +8,15 @@ def __init__(self, name):
 class Colors():
```
tessbianchi: FANCY PRINT!

utils/navigator_tools/nodes/fake_action_server.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
```
tessbianchi: BEAUTIFUL!

PR #148, opened 2016-11-19
Lidar Improvements

maverick7170:
Lots of things going on this PR but here are the important points:

- Judge's GUI is added to the utilities folder (this isn't lidar specific code but whatever....)
- reset (for objects) and Reset (for ROI) cmds added to reset while online
- improved object tracking by changing height reference point and adding a confidence parameter to lidar data and classification data
- after inflating ogrid and finding obstacles, the ogrid is eroded (deflated) to help with driving performance (specifically the dock)
- bags from lake day 11/12 were used to verify object classification and all objects classified correctly (people on kayaks still cause issues and the dock is messy but works). We just need to document what cases the classifier doesn't work this weekend so I can compare to my test data to find the issue.

Tomorrow we should be able to quantify how well/poorly this new setup will work!


ironmig:
Fixed the build issues and pushed to new PR #149

PR #149, opened 2016-11-19
JUDGES GUI, OBJECT CLASSIFICATION: Ira changes from last few weeks

ironmig:
Lots of things going on this PR but here are the important points:

- Judge's GUI is added to the utilities folder (this isn't lidar specific code but whatever....)
- reset (for objects) and Reset (for ROI) cmds added to reset while online
- improved object tracking by changing height reference point and adding a confidence parameter to lidar data and classification data
- after inflating ogrid and finding obstacles, the ogrid is eroded (deflated) to help with driving performance (specifically the dock)
- bags from lake day 11/12 were used to verify object classification and all objects classified correctly (people on kayaks still cause issues and the dock is messy but works). We just need to document what cases the classifier doesn't work this weekend so I can compare to my test data to find the issue.

Tomorrow we should be able to quantify how well/poorly this new setup will work!


perception/navigator_lidar_oa/src/VolumeClassifier.h
```diff
@@ -11,51 +11,57 @@
...
-	//If we classify something as shooter/scanthecode, leave it alone
+
+	//Skip classification if the object is locked, set as a start_gate, or if it isn't real
 	if (object.locked || object.name == navigator_msgs::PerceptionObject::START_GATE_BUOY || !object.real) {
 		return;
 	}
 
+	//Possible classifications
 	std::vector<std::string> names = {"dock","shooter","scan_the_code","totem","buoy"};
-	double volumes[5][8] = { 	{0.7,	1.25,	6.0,	7.0,	6.0,	7.0,	2.5,	3.25}, //dock (NOT TESTED!)
```
tessbianchi: @maverick7170  

Can you put a comment about what these number means so that we can edit them in Hawaii?

PR #154, opened 2016-11-22
CAMERA: Calibration for downwards camera

RustyBamboo:


mission_control/navigator_launch/launch/subsystems/cameras/calibration/00b09d01009ce76c.yaml
```diff
@@ -0,0 +1,20 @@
+image_width: 752
+image_height: 480
+camera_name: narrow_stereo
```
mattlangford: Is this name important?

PR #155, opened 2016-11-23
LIDAR: classification fixes from lakeday 11/19-11/20

ironmig:
@maverick7170's changes to classification from lakeday 11/19 through 11/20.


perception/navigator_lidar_oa/src/objects.h
```diff
@@ -205,6 +206,7 @@ class ObjectTracker
 				thisOne.pclInliers = s_obj.pclInliers;
```
tessbianchi: header.stamp should not be s_obj.age

tessbianchi: Also, I am using strikesFrame right now, and it is not returning any points.

PR #160, opened 2016-11-30
REMOTE_CONTROL: Add emergency controller nodes 

zobeld:
This adds two nodes and firmware for the emergency controller in /utils/remote_control. They listen to a receiver board through usb and publish data to wrench/cmd in a way that mimics the xbox controller. Most of the code is butchered from people who knew what they were doing, so it may be too inelegant to be reviewed. I'm not a CE and have no idea what I'm doing with git, and now's a terrible time to start, so please delete this if it's not up to standard or sufficiently organized.

There are two nodes added in /remote_control/navigator_emergency_control:
- The emergency_control.cpp node is retooled from @forrestv's stm32f3discovery_imu_driver code in hardware-common, in order to interact with a uP on a receiver board. This node publishes data received by the xbee in /joy_emergency, similar to the xbox_joy_node.
- The emergency_control.py node copies @whispercoros's navigator_joystick node and subscribes to /joy_emergency, publishes to /wrench/emergency. This is added to the wrench arbiter, and publishes to /wrench/cmd in the same way as the xbox controller, when selected or toggled to.

Most of the files added are from the hardware-common repository for loading firmware onto the stm32f3, including libopencm3. The important files outside of /remote_control/navigator_emergency_control that are changed are:
- wrench_arbiter: added "emergency" input
- remote_control_lib: added a function to select emergency controller as active controller
- gnc and simulation launch files: launch the c++ and py nodes
- install.sh: added dependency installations for libopencm3, according to the readme for hardware-common


ironmig:
Looks like the build is failing because hardware-commons has a submodule which isn't getting cloned by the install script. Try adding this
```
diff --git a/install.sh b/install.sh
index c00f9f6..712b8eb 100755
--- a/install.sh
+++ b/install.sh
@@ -77,7 +77,7 @@ ros_git_get() {
 	done
 	if $NEEDS_INSTALL; then
 		instlog "Installing $INSTALL_URL in $CATKIN_DIR/src"
-		git clone -q $INSTALL_URL --depth=1
+		git clone -q $INSTALL_URL --depth=1 --recursive
 	fi
 }
```

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
+
+
+from remote_control_lib import RemoteControl
+import rospy
+from sensor_msgs.msg import Joy
+
+
+__maintainer__ = "Anthony Olive"
```
ironmig: I think you can be the author here!

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
```diff
@@ -0,0 +1,124 @@
...
+    def reset(self):
+        '''
+        Used to reset the state of the controller. Sometimes when it
+        disconnects then comes back online, the settings are all out of whack.
+        '''
+        self.last_kill = False
+        self.last_station_hold_state = False
+        self.last_auto_control = False
+        self.last_emergency_control = False
+        self.last_shooter_load = False
```
ironmig: Can you add shooter_cancel and get rid of load and fire here? I imagine with the emergency controller we'd only ever really want to cancel the shooter incase someone's hand was there or something.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
zobeld: Ah, I was wondering what I should do about that. I only changed one line of code so I thought Anthony should be kept as author. But I also thought if something goes wrong with the node, my name should be there so Anthony can have deniability. I'll go with your suggestion if that's standard.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
zobeld: Roger, will do. There were extra buttons on the controller, so I just added anything. If there are any other functions that would be more useful, like set to keyboard or toggle state, I'll replace those too.

zobeld:
This has been tested on the simulator, but not with the boat itself. From what I've seen, it won't interfere with normal operation as long a select_emergency_control isn't called, or if another controller toggles modes to it. If the controller stops sending data or the firmware on the receiver crashes for whatever reason, the c++ node times out and zeros everything.

And sure, I have nothing going on in my life, ever, so I can be there on Sunday.

The hardware is just a receiver board with firmware to interpret data from the xbees. It connects through usb and has a small 10" antenna. But I couldn't get the go-ahead yesterday to install it in the box. With the hardware not plugged in, there's just an error message failing to open the port, nothing interferes as far as I can tell.

I'll put together the readme later today. Thanks for the advice!

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
whispercoros: According to precedent, the author of the file is it's original creator, so I agree with Kevin here.

utils/remote_control/navigator_emergency_control/navigator_emergency_control/mainpage.dox
```diff
@@ -0,0 +1,14 @@
+/**
+\mainpage
+\htmlinclude manifest.html
+
+\b navigator_emergency_control 
+
+<!-- 
+Provide an overview of your package.
```
whispercoros: This documentation file is pretty barren, could you please either add a bit more to it or remove it? Your readme file was fairly comprehensive, so I think this can be removed.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+rospy.init_node("emergency")
+
+
+class Joystick(object):
+
+    def __init__(self):
+        self.force_scale = rospy.get_param("~force_scale", 600)
+        self.torque_scale = rospy.get_param("~torque_scale", 500)
```
whispercoros: The ~ in lines 21-22 refers to parameters set in the namespace of the node (which would take place in the launch file). These are both defined in the /joystick_wrench namespace, so that must be accounted for here.

        self.force_scale = rospy.get_param("/joystick_wrench/force_scale", 600)
        self.torque_scale = rospy.get_param("/joystick_wrench/torque_scale", 500)

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
```diff
@@ -0,0 +1,137 @@
...
+
+	    # No change in state
+	    if rospy.Time.now() - self.last_joy.header.stamp > rospy.Duration(2):
+
+	        # No new instructions after 2 seconds, zero the wrench
+                self.reset()
+
+if __name__ == "__main__":
+    emergency = Joystick()
+    rospy.Timer(rospy.Duration(1), emergency.timeout, oneshot=False)
```
whispercoros: Your docstring for the timeout function states that a timeout will occur after 2s; however, this timer is set to go off on a 1s interval.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
```diff
@@ -0,0 +1,137 @@
...
+        self.last_shooter_cancel = False
+        self.last_change_mode = False
+
+        self.start_count = 0
+        self.last_joy = None
+        self.active = False
+
+        self.remote.clear_wrench()
+
+    def check_for_timeout(self, joy):
```
whispercoros: This function is replaced by your timeout function, so I would recommend removing it and all reverences to it.

install.sh
```diff
@@ -373,6 +373,10 @@ sudo apt-get install -qq git-lfs gitk
 git lfs install --skip-smudge
 sudo apt-get install -qq tmux
 
+# Tools needed for hardware-common
+sudo mkdir -p /etc/apt/preferences.d/ && sudo bash -c "echo -e 'Package: *\nPin: origin "ppa.launchpad.net"\nPin-Priority: 999' > /etc/apt/preferences.d/arm" && sudo rm -f /etc/apt/sources.list.d/terry_guo-gcc-arm-embedded-*.list && sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded && sudo apt-get update -qq && sudo apt-get remove -qq -y gcc-arm-none-eabi binutils-arm-none-eabi libnewlib-arm-none-eabi libnewlib-dev && sudo apt-get install -qq -y --force-yes gcc-arm-none-eabi
```
zobeld: Roger, removed the &&s, but the sudo bash command I don't think I can shorten, it's around 120 chars.

mission_control/navigator_launch/launch/simulation.launch
```diff
@@ -38,6 +38,11 @@
         <param name="autorepeat_rate" value="5"/>
     </node>
 
+    <node pkg="navigator_emergency_control" type="navigator_emergency_control" name="navigator_emergency_xbee">
```
zobeld: Ah, I didn't fully understand which files the controller node should launched in. I tried to add it in the same locations as the node for the xbox controller, and that was simulation and master_shore. I figured the shore wouldn't launch with the boat itself, so I didn't add it there, but I guess I forgot to actually put it somewhere it'll launch with the boat. I think the most appropriate place would be hardware_drivers, so I'll add it there.

I've been using the simulator to test the output of wrench/cmd. Ever since I updated to dev a while ago, the simulator didn't run properly, but I wasn't sure if there was just something wrong with Gazebo locally. Are you saying the simulator is malfunctioning for everyone?

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
zobeld: Oh thanks, I had thought it would grab these parameters from /joystick_wrench with this notation. I guess it only worked because it used the default value.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
zobeld: Well, my timeout function is supposed to only check if the hardware disconnects or associated node dies and stops publishing. I still want the controller to timeout after 15 minutes of no change.

The difference is, if the node dies, /joy_emergency stops publishing completely, and it should use my timeout to see that. If the controller just isn't touched, and the node is still alive, /joy_emergency is still publishing zeros, and shouldn't activate my timeout. It should still activate your timeout after 15 minutes if the controller doesn't change state.

On closer inspection, it looks like my function didn't behave the way I just described, so I reworked it to look at a new variable last_time, which is the last time of a received /joy_emergency. Also renamed it die_check.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
zobeld: Yes, but the duration within the function is set to 2s. If the timer were, say, 1ms, it would check every ms if the time interval between now and the last received joy is greater than 2s. As it is right now, it does that every second, so I suppose it's really activating between 2-3s. Setting the timer to 2s would be anywhere between 2-4s. I could make it more accurate by making the timer a smaller duration, but I don't think there's any need to run it very often for the sake of accuracy.

mission_control/navigator_launch/launch/simulation.launch
mattlangford: Any teleop control in the sim is disabled in favor of perfect tracking for mission testing. 
The best way to test funcationality is to look at `wrench/cmd` and make sure that is doing what you want it to.
There's some changes that could be made to the sim to let teleop be added back in if someone wants to do that as well.

whispercoros: I think as long as `/wrench/cmd` is outputting the correct values, this should work. All of the buttons use functions in the remote control library, which have been throughly tested. I would recommend placing the launcher for whatever needs to be running on the boat in `gnc.launch` as opposed to hardware_drivers, this is where joystick and keyboard control are brought up. While you're at it, joystick control should be removed from `simulation.launch` as it is no longer used there.

mission_control/navigator_launch/launch/simulation.launch
```diff
@@ -38,6 +38,11 @@
...
+        <param name="port" type="string" value="/dev/serial/by-id/usb-uf-mil_xbee_b267017f-if00" />
```
whispercoros: Sorry for the confusing wording. What I meant to ask was if using the chip on the boat or any of our computers would give the same serial ID. It sounds like this is the case.

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
whispercoros: Good idea, sorry I missed that nuance in my quick read through.

install.sh
whispercoros: 120 is an okay length if it cannot be shortened. I didn't see this change committed, so please add it to the next commit.

utils/remote_control/navigator_emergency_control/navigator_emergency_control/mainpage.dox
whispercoros: Has anything been done with this file yet?

utils/remote_control/navigator_emergency_control/nodes/nodes/navigator_emergency.py
whispercoros: Sounds good

whispercoros:
Still waiting on @ironmig to approve his requested changes. He should merge this when he is done.

PR #164, opened 2016-12-04
GUI and REMOTE: Dashboard and remote control fixes

whispercoros:
I have fixed the long-standing segfault errors that the dashboard used to have. These were likely due to the fact that I originally wrote it thinking that ROS timers were fully multi-threaded, whereas they are serialized. The fix consists of offloading the task that caused the vast majority of blocking to a separate node (/host_monitor).

I also added links to the new emergency controller in each of our current control interfaces (figured better safe than sorry on this one). Obviously, this means I have to merge #160 and rebase this to dev before it is merged, but I wanted to get the review process started early.


utils/navigator_msgs/msg/Hosts.msg
```diff
@@ -0,0 +1 @@
+string[] hosts
```
ironmig: You should take more  advantage of ros msgs here. Ros msgs can be maps (dictionaries), so it isn't so unclear what the format of the data in this message is.

utils/remote_control/navigator_joystick_control/nodes/navigator_joystick.py
```diff
@@ -34,10 +34,11 @@ def reset(self):
         '''
         self.last_kill = False
         self.last_station_hold_state = False
-        self.last_change_mode = False
-        self.last_auto_control = False
         self.last_rc_control = False
+        self.last_emergency_control = False
```
ironmig: Note: Do not merge this until #160 is merged or errors.

utils/navigator_gui/package.xml
```diff
@@ -4,6 +4,7 @@
   <version>1.0.0</version>
   <description>The navigator_gui package</description>
 
+  <author email="anthony@iris-systems.net">Anthony Olive</author>
```
ironmig: ###### k

utils/navigator_gui/navigator_gui/host_monitor.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class HostMonitor():
+
+    def __init__(self):
+        self.pub_hosts = rospy.Publisher("/host_monitor", Hosts, queue_size=1)
+
+        # Build an ordered list of host dictionaries that resolve to devices on navigator
+        host_list = [
+            "mil-nav-ubnt-wamv",
```
ironmig: What machines do these hostnames map to? There are three for wamv. It would also be nice to add the IP of the Ubiquity on both sides, as it would be very helpful for debugging network outages.

utils/navigator_msgs/msg/Hosts.msg
ironmig: Also you should define the hostname constants here for use in the check hosts nodes to follow ros convention

In message file
```
string WAMV=mil-nav-wamv
```
In code
```
checkHost(Hosts.WAMV)
```
etc

utils/navigator_gui/navigator_gui/dashboard.py
```diff
@@ -6,11 +6,11 @@
 '''
 
 
+import functools
 import os
-import socket
-import subprocess
```
ironmig: damn right

mission_control/navigator_launch/launch/subsystems/sensors.launch
```diff
@@ -7,4 +7,7 @@
 
   <!-- Sick -->
   <include file="$(find navigator_launch)/launch/subsystems/sick_lidar.launch"/>
+  
+  <!-- Host Monitoring -->
+  <node name="host_monitor" pkg="navigator_gui" type="host_monitor.py">
```
ironmig: I think this should actually be in master shore launch. Whoever runs network kill then will also be checking that everything is connected. We don't want the boat running this, because we won't be able to see what is connected if the boat is not.

utils/navigator_msgs/msg/Hosts.msg
whispercoros: Thanks for the tip, this should help cut down on the code I was using to pass the dictionary.

utils/navigator_gui/navigator_gui/host_monitor.py
whispercoros: mil-nav-ubnt-wamv = The Ubiquity on NaviGator
mil-nav-ubnt-shore = The Ubiquity on shore
mil-nav-wamv = The primary computer
mil-nav-perception = The secondary computer
mil-com-velodyne-vlp16 = The Velodyne LiDAR
mil-com-sick-lms111 = The Sick LiDAR

utils/navigator_gui/navigator_gui/dashboard.py
whispercoros: I should point out that these were just moved to the host monitor, not removed all together.

mission_control/navigator_launch/launch/subsystems/sensors.launch
whispercoros: Does only one person usually run `master_shore.launch`? Multiple people trying to run the node would be unnecessary. This makes sense if that is the case.

ironmig: Only one person will run master shore, whoever does network kill and joystick

ironmig:
The emergency control was merged, so as soon as you do the host checker refactors I suggested we are good to merge.

whispercoros:
This is good to merge once it builds.

mattlangford:
Same as Kevin's PR, I'll build and merge

PR #171, opened 2016-12-05
DETECT DELIVER: Shoot in random shape if correct not found

ironmig:
- As detect deliver runs, it remembers what shapes it saw. If after circling around the platform the correct shape/color is not identified, it will shoot in one of the targets it saw while circling
- Catch some potential edge case errors in CameraLidarTransformer
- Update the Searcher class in navigator to take a custom looking callback. I don't think anyone else uses this, so it won't break anything.
- Everything has been tested in bags without any regressions

### DIFF
--- a/mission_control/navigator_missions/nav_missions/detect_deliver.py
+++ b/mission_control/navigator_missions/nav_missions/detect_deliver.py
@@ -33,6 +33,11 @@ class DetectDeliverMission:
             self.navigator.nh, '/shooter/load', ShooterDoAction)
         self.shooterFire = txros.action.ActionClient(
             self.navigator.nh, '/shooter/fire', ShooterDoAction)
+        self.identified_shapes = []
+        self.last_shape_error = ""
+        self.last_lidar_error = ""
+        self.found_shape = None
+        self.normal_res = None
 
     def _bounding_rect(self,points):
         np_points = map(navigator_tools.point_to_numpy, points)
@@ -66,9 +71,8 @@ class DetectDeliverMission:
         pattern = self.navigator.move.d_circle_point(navigator_tools.rosmsg_to_numpy(
             self.waypoint_res.objects[0].position), radius=self.circle_radius, theta_offset=self.theta_offset, direction='cw')
         yield next(pattern).go()
-        searcher = self.navigator.search(
-            vision_proxy='get_shape', search_pattern=pattern, Shape=self.Shape, Color=self.Color)
-        yield searcher.start_search(timeout=self.search_timeout_seconds, spotings_req=self.spotings_req, move_type="skid")
+        searcher = self.navigator.search(search_pattern=pattern, looker=self.search_shape)
+        yield searcher.start_search(timeout=self.search_timeout_seconds, spotings_req=self.spotings_req, move_type="skid", loop=False)
         print "Ended Circle Search"
 
         #  fprint("Starting Circle Search", title="DETECT DELIVER",  msg_color='green')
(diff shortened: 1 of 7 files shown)


mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -79,58 +82,84 @@ def circle_search(self):
         #  yield self.navigator.move.stop()
         #  fprint("Ended Circle Search", title="DETECT DELIVER",  msg_color='green')
 
+    def update_shape(self, shape_res, normal_res, tf):
+       self.identified_shapes[(shape_res.Shape, shape_res.Color)] = self.get_shape_pos(normal_res, tf)
+
+    def correct_shape(self, shape):
+        return (self.Color == "ANY" or self.Color == shape.Color) and (self.Shape == "ANY" or self.Shape == shape.Shape)
```
mattlangford: Use a Karanugh Map and simplify this
/s

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -79,58 +82,84 @@ def circle_search(self):
...
     @txros.util.cancellableInlineCallbacks
-    def align_to_target(self):
-        self.markers = MarkerArray()
-        while True:
-            if (yield self.is_found()):
-                if (yield self.get_normal()):
-                    break
+    def search_shape(self):
+        shapes = yield self.get_shape()
+        if shapes.found:
```
mattlangford: Navigator checks for this by default, you probably don't want that exception to be thrown right,

mission_control/navigator_missions/navigator_singleton/navigator.py
```diff
@@ -332,13 +332,17 @@ def _valid(self,value):
...
 class Searcher(object):
-    def __init__(self, nav, vision_proxy, search_pattern, **kwargs):
+    def __init__(self, nav, search_pattern, looker=None, vision_proxy="test", **kwargs):
         self.nav = nav
-        self.vision_proxy = vision_proxy
-        self.vision_kwargs = kwargs
+        self.looker = looker
+        if looker == None:
+            self.looker = self._vision_proxy_look 
+            self.vision_proxy = vision_proxy
```
mattlangford: Do these two statements need to be reordered?

mattlangford: I guess it's okay since `self.looker` doesn't get called until later

mattlangford:
I'm going to pull this on my computer and try building. When it works I'll go ahead and merge

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -79,58 +82,84 @@ def circle_search(self):
...
-        self.enupoint = enu_cam_tf.transform_point(navigator_tools.rosmsg_to_numpy(self.normal_res.closest))
-        defer.returnValue(True)
+        normal_res = yield self.cameraLidarTransformer(req)
+        if not self.normal_is_sane(normal_res.normal):
+            normal_res.success = False
+            normal_res.error = "UNREASONABLE NORMAL"
+        defer.returnValue(normal_res)
 
     def normal_is_sane(self, vector3):
          return abs(navigator_tools.rosmsg_to_numpy(vector3)[1]) < 0.2
```
tessbianchi: Small change, but it would probably be better to put any threshold values like this at the top of the file in a variable, just in case you want to change them later. 

However, if you think this value will be fairly constant, then it is fine.

mission_control/navigator_missions/nav_missions/detect_deliver.py
```diff
@@ -79,58 +82,84 @@ def circle_search(self):
...
+            self.last_shape_error = shapes.error
+        defer.returnValue(False)
+
+    def select_backup_shape(self):
+        for (shape, color), point_normal in self.identified_shapes.iteritems():
+            self.shape_pose = point_normal
+            if self.Shape == shape or self.Color == color:
+                fprint("Correct shape not found, resorting to shape={} color={}".format(shape, color), title="DETECT DELIVER",  msg_color='yellow')
+                return
+        if self.shape_pose == None:
```
tessbianchi: If you haven't recognized any shape, do you still want to fire, or no?

PR #172, opened 2016-12-06
PERCEPTION, USI

DSsoto:
* Underwater Shape Identification node and UnderwaterShapeDetector class
* added an image filtering module to navigator_vision_lib
* add yaml config for USI
* support for automated loading of shape templates with creation of corresponding Shape classes

Some of the convolutional methods I added here will not work well for USI since I made the assumption that the shapes would be at a constant depth. I asked on the RobotX group and that is not the case. I am working on an alternative method, however the new module can be very useful for a wide variety of vision tasks.

@ironmig I know you think that an uncaught exception is the best way to exit a program but I changed that for tf_to_gazebo


utils/navigator_tools/include/navigator_tools.hpp
```diff
@@ -5,6 +5,9 @@
 #include <ros/ros.h>
 
 namespace nav{
+
+static const double PI = 3.1415926535897932;
```
DSsoto: This isn't python. The alternative is: 
```
"Many C++ compilers define M_PI in cmath (or in math.h for C) as a non-standard extension.
You may have to #define _USE_MATH_DEFINES before you can see it."
```

perception/navigator_vision/nodes/underwater_shape_identification_vision.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <missions/underwater_shape_identification.hpp>
+
+using namespace std;
+
+int main(int argc, char** argv) 
+{ 
+  cout << "\033[1;31mUnderwater Shape Identification Vision\033[0m" << endl;  
+
+  // set up image acquisition
+  string cam_topic {"down/image_raw"};
```
DSsoto: Will do.

DSsoto:
The build failed because of an internal compiler error, you can look at the build. It builds fine on mine. 

I didn't claim to have the node interface completed, just wanted to PR incrementally. The image filtering code has been tested and works correctly. I had to switch the approach I was using because of a bad assumption as I explained above, service interface will come in the next PR probably.

ironmig:
I'm still a little worried that the build is still failing after trying again, while all the other PR's aren't failing, but whatever. I'm all about incremental PR's so no prob there. Please ```pull --rebase``` and we good.

DSsoto:
Done, build failed twice due to compiler errors but succeeded in build 3. Weird.

ironmig:
don't think you rebased right. Looks like you rebased dev on top of your branch, not the other way around. Try::
```
git checkout 377d01c26f6
git checkout -b fix-stuff
git pull --rebase upstream dev
git push YOURREMOTE fix-stuff:dev2
```

DSsoto:
My bad, fixed.

PR #181, opened 2016-12-13
KILL, TOTEMS, OGRID

mattlangford:
Lot's of updates for the kill board that are being tested and need to be on the boat.

I switched to a different, more reliable way to do circles and added the database colorizer to the perception launch file. There is code in there to test the circling as well.

Added a way to replace portions of the ogrid using a new set of topics as specified in the dynamic reconfig. On that note, I added bounds for the channel we are testing in


mission_control/navigator_missions/nav_missions/circle_tower.py
```diff
@@ -39,7 +45,7 @@ def main(navigator, **kwargs):
     # TODO: Check which side to go to based on relative distance
     center = navigator.move.set_position(buoy_field_point).set_orientation(get_solar_q())
     obs_point = center.backward(BF_WIDTH / 2 + BF_EST_COFIDENCE) 
-    yield obs_point.go()
+    #yield obs_point.go()
```
mattlangford: Lot's of commented out code here since we will want to test the main functionality first before testing the whole mission

PR #184, opened 2016-12-17
GREEN DOT OPPORTUNITY

DSsoto:
* Made distance b/w observation points smaller
* Fixed bad return from ros service
* Active contours


ironmig:
Included with fixes in #186

PR #191, opened 2016-12-20
Final master dev merge

mattlangford:


whispercoros:
We won, so this sort of worked.

PR #201, opened 2017-08-30
KINETIC: changes for 16.04/kinetic compatiblity and mil_common integration

ironmig:
* Make changes necessary to maintain kinetic compatibility 
* Remove files/packages redundant with mil_common replacements
* Replace navigator alarm system with ros_alarms
* Create deprecated directory (not built with catkin) that contains old, unused code that we might want to keep around for a while
* Remove/Deprecate old, unused or broken code
* Disable building of navigator_emergency_controller... I have not yet seen this in action and it presents a lot of problems for building with ARM. If someone wants to try to make this better, we can reenable it
* NOTE: will only build/run with uf-mil/mil_common#82

There are a LOT of changes in here, and I may have missed some bugs. I tested in the simulator and was able to get Navigator to move (a lot of things have to work for this to happen). But for now I'm just PRing to a feature branch that we can work on before officially switching master to kinetic


RustyBamboo:
Let's just merge this to add some convince to testing

Grymestone:
Merging this will force us to actually start testing it so let's just merge. #Approved.

PR #213, opened 2017-09-04
PATH PLANNER: clone lqrrt node into repo

ironmig:
I didn't like how we were handling our use of lqrRRT, where we just run the ROS demo in that repository (which may change and not be inline with NaviGator specific needs). It also requried us manually run a setup.py as lqRRT is not a catkin package. This addresses both of those problems, by cloning the current lqRRT ROS Demo into the repository so we can develop it in NaviGator in the future (or replace it with a different path planner).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "gnc/lqRRT"]
-	path = gnc/lqRRT
-	url = https://github.com/jnez71/lqRRT
+	path = gnc/navigator_path_planner/lqRRT
+	url = https://github.com/uf-mil/lqRRT
--- a/gnc/navigator_path_planner/package.xml
+++ b/gnc/navigator_path_planner/package.xml
@@ -1,0 +1,19 @@
+<package format="2">
+
+  <name>navigator_path_planner</name>
+  <version>0.0.0</version>
+  <description>Node for NaviGator's lqRRT based path planner</description>
+  <maintainer email="nez.jason7@gmail.com">jason</maintainer>
+  <license>MIT</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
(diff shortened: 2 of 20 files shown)


.gitmodules
```diff
@@ -1,6 +1,6 @@
-[submodule "gnc/lqRRT"]
-	path = gnc/lqRRT
-	url = https://github.com/jnez71/lqRRT
+[submodule "gnc/lqrrt"]
+	path = gnc/navigator_path_planner/lqrrt
```
RustyBamboo: `path = gnc/navigator_path_planner/lqRRT`

RustyBamboo: The path currently is `gnc/navigator_path_planner/lqRRT`

gnc/navigator_path_planner/package.xml
```diff
@@ -0,0 +1,19 @@
...
+  <name>navigator_path_planner</name>
+  <version>0.0.0</version>
+  <description>Node for NaviGator's lqRRT based path planner</description>
+  <maintainer email="nez.jason7@gmail.com">jason</maintainer>
+  <license>MIT</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <depend>rospy</depend>
+  <depend>actionlib</depend>
+  <depend>actionlib_msgs</depend>
```
RustyBamboo: Need build_depend and run_depend for actionlib?

RustyBamboo:
Replace lqrrt_ros.msg in `mission_control/navigator_alarm/navigator_alarm_handlers/station_hold.py`

ironmig:
Good catch... I think I'm actually going to PR this to kinetic-upgrade because that's what everyone's working on

PR #216, opened 2017-09-05
KILL BOARD: refactor driver, add simulated serial mode

ironmig:
I refactored the kill board driver, as it was poorly documented and messy. I also added a simulated kill board class, which will operate on the kill board's protocol so we can test this code without physical access to the board


ironmig:
This is tested and works on NaviGator, at least in terms of knowing when the board is killed and passing along ros_alarms kills to the board. Individual feedback for each button or the remote kill is not reliable because the board does not correctly send updates when these events happen (verified by looking at the serial bytes coming in from the board as we press buttons). I might try to work around this by periodically requesting the status of each button..

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
```diff
@@ -0,0 +1,221 @@
...
+
+        self.hw_kill_broadcaster = AlarmBroadcaster('hw-kill')
+
+        AlarmListener('hw-kill', self.hw_kill_alarm_cb)
+        AlarmListener('kill', self.kill_alarm_cb)
+        rospy.Subscriber("/wrench/current", String, self.wrench_cb)
+        rospy.Subscriber("/network", Header, self.network_cb)  # Passes along network hearbeat to kill board
+
+    def connect(self):
+        if rospy.get_param('/is_simulation', False):  # If in Gazebo, run fake serial class following board's protocol
```
RustyBamboo: This kind of info should probably go into a readme

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
```diff
@@ -0,0 +1,221 @@
...
+        # If a response to another request, like ping or computer kill/clear is recieved
+        for index, byte in enumerate(self.expected_responses):
+            if msg == byte:
+                del self.expected_responses[index]
+                return
+        # Log a warning if an unexpected byte was recieved
+        rospy.logwarn('Recieved an unexpected byte {}, remaining expected_responses={}'.format(
+            hex(ord(msg)), len(self.expected_responses)))
+
+    @thread_lock(lock)
```
RustyBamboo: some voodoo

ironmig:
Tested extensively in CIMAR, works very reliably.

PR #218, opened 2017-09-07
CI: Add Jenkins and Dockerfile

Grymestone:
These are the files necessary to get the CI working. Once kinetic-upgrade is deployed this will be deployed alongside it and I can build the image on plumbusi. Until then, trying to force this in before cmake works will only cause problems with existing CI operations. 

This is mainly due to the fact both Jenkins and Docker rely on the install script which still points to the master branch. Otherwise there may be some issues with the SDK for point gray, but that should be a near instant fix if said issue crops up. It doesn't seem like it will, thanks to the install script, but it is worth noting.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,46 @@
+dockerNode(image: 'uf-mil:navigator') {
+	stage("Checkout") {
+		checkout scm
+		sh '''
+			source ~/.mil/milrc > /dev/null 2>&1
+			git submodule update --init --recursive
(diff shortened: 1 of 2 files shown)


ironmig:
Looking at the install script, if it running under docker it does not attempt to build the repo. Is there some error preventing the image from being built? It looks like Jenkins is correctly trying to build it now!

ironmig:
I built the image. It is now correctly building (though failing a test which I will now look into). Great work on this!

PR #228, opened 2017-09-14
FORMAT: auto format C++ files

ironmig:


RustyBamboo:
This passed the building so should be all good

PR #233, opened 2017-09-19
GUI: fix Qt API change in 16.04

ironmig:
Can someone else verify that that navigator GUI works on 16.04 with this change? Build, open rqt, and add the NaviGator gui as a plugin (suggest running simulation first)

Edit: I just mean works initially like you can see the interface and buttons. It might freeze or segfault after some time as it has always done...

### DIFF
--- a/utils/navigator_gui/navigator_gui/dashboard.py
+++ b/utils/navigator_gui/navigator_gui/dashboard.py
@@ -1,24 +1,19 @@
 #!/usr/bin/env python
-
 '''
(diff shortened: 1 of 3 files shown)


RustyBamboo:
Works for me. Any reason you didn't modify the shooter.py or alarms?
screenshot from 2017-09-28 18-46-52

ironmig:
Fixed the shooter plugin too and refactored the dashboard a bunch. It should not crash anymore (let me know if it does)

PR #251, opened 2017-10-20
BOOT: clean up boot process

ironmig:
* Add special sause commands to run Sylphase hardware interface into repo, will spawn in tmux along with roscore
* Get rid of old on_boot package, it just didn't make no sense
* Import the little json decoder to get the solutions from the Sylphase software into ROS so we can run it with the rest of our system and change it to our likeing (for example, transforming it into our base_link frame in the future)


scripts/on_boot
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
RustyBamboo: lol

ironmig:
Believe it or not this still doesn't work! If I run rc.local as root myself, it's fine. But when it's run automatically by the system it still prompts for password! Idk

RustyBamboo:
oh geeez. but eh, good for testing.

sub8:
Try adding the navigator user to the sudoers file with the NOPASSWD option

mission_control/navigator_alarm/navigator_alarm_handlers/kill.py
```diff
@@ -45,8 +45,8 @@ def meta_predicate(self, meta_alarm, alarms):
         ignore = []
 
         # Don't kill on low battery, only on critical
-        if alarms['battery-voltage'].raised and alarms['battery-voltage'].severity < 2:
-            ignore.append('battery-voltage')
+        #if alarms['battery-voltage'].raised and alarms['battery-voltage'].severity < 2:
```
RustyBamboo: Not sure we should change this..

ironmig: Well right now it's broken because it uses voltage readings from killed thrusters. There is issue #254 to fix this but for now we should not use it

PR #261, opened 2017-10-26
Update master branch to kinetic-upgrade

ironmig:
With the PRs just merged, we now have code that ran a working system at the last test day. I think it is time to set our master branch to officially run on kinetic!


ironmig:
Also nice to note that in total this removes 5000 lines from our repo while maintaining same functionality! And I bet if we exclude deprecated it's way more

ironmig:
I think it's better to merge this so we don't confuse the new people. Mostly everything is working from the last two test days (with bounds being the major exception)

PR #265, opened 2017-10-26
BATTERY: ignore voltage readings if fault

RustyBamboo:


ironmig:
I went over to CIMAR with David for kill testing (buttons work again woo!) and did some more verification of how the motor controllers respond to kills / off batteries. Heres what I learned:

* When the battery powering a thruster is off (front for left battery, rear for rear battery), the thruster reports the undervoltage fault
* When thrusters are being commanded a thrust, undervault fault is reliably raised on kill and not raised on revive
* When a thruster is not being commanded a thrust or is commanded a very small thrust (<10), undervault fault IS NOT raised when we are killed. The voltage reading becomes inaccurate and trends towards zero, but the fault is not seen

So this probably means that the fault code won't be that reliable for detecting kill. I think the best way is to monitor hw-kill alarm and ignore all readings if it is raised. While you're there though, maybe we should create a new thruster-fault alarm that raises when a thruster fault is engaged with the parameters having a nice string decoding of the byte

RustyBamboo:
yay hardware

PR #270, opened 2017-10-30
MISSIONS: integrate mil_tasks

ironmig:
**depends on uf-mil/mil_common#87**

Rework of navigator_missions package to use the new mil_tasks package. For now, the old mission planner is deprecated until it is refactored to work with mil_tasks. Missions can now be run from the mil_tasks GUI or ```runtask MISSIONNAME``` or ```nmove MOVECOMMAND``` aliases. Some skeleton missions that did nothing are also deprecated. It is possible that some missions were broken in the refactor, but I was careful to preserve their behavior.

TODO:
- [ ] test move command in sim
- [ ] replace all prints/fprints with send_feedback
- [ ] test other refactored commands


mission_control/navigator_missions/launch/task_runner.launch
```diff
@@ -0,0 +1,6 @@
+<launch>
```
ironmig: Probably rename to missions.launch

ironmig:
replaced by #274

PR #271, opened 2017-10-31
ALARMS: Add thruster alarm that raises on fault

RustyBamboo:


mission_control/navigator_alarm/navigator_alarm_handlers/thruster_fault.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _check_faults(self, msg, topic):
+        if msg.fault == 0:
+            return
+
+        # Figure out what fault codes have been raised and find their meaning from dictionary
+        get_fault_codes = []
+        for i in range(0, len(self.fault_codes)):
+            # Fault message is the sum of binary strings
+            decode = msg.fault & 2**i
```
ironmig: Shouldn't this just be
```
for key in self.fault_codes:
    decode = msg.fault & key
```
I don't think you need to raise 2 to i. 2^128 is way beyond integer limit

RustyBamboo: I guess that's the same thing cause 2^i gives 0,1,2,...,128. Using key is a little more efficient i suppose

mission_control/navigator_alarm/navigator_alarm_handlers/thruster_fault.py
```diff
@@ -0,0 +1,50 @@
...
+            decode = msg.fault & key
+            if decode != 0:
+                get_fault_codes.append(value)
+
+        if not self._raised:
+            self.broadcaster.raise_alarm(
+                severity=5,
+                problem_description='{} has faults: {}'.format(
+                    topic, ' '.join(get_fault_codes)),
+                parameters={'fault': msg.fault}
```
ironmig: Should report if multiple thrusters have faults, so needs to internally track each thrusters fault then raise alarm if there are any changes. The list of faults by thruster should be in the parameters so other programs can decode them as json instead of in the problem description string

PR #274, opened 2017-11-14
MISSIONS: integrate mil_tasks

ironmig:
**used to be #270, re-PRing to go to master branch**

**depends on uf-mil/mil_common#87**

Rework of navigator_missions package to use the new mil_tasks package. For now, the old mission planner is deprecated until it is refactored to work with mil_tasks. Missions can now be run from the mil_tasks GUI or ```runtask MISSIONNAME``` or ```nmove MOVECOMMAND``` aliases. Some skeleton missions that did nothing are also deprecated. It is possible that some missions were broken in the refactor, but I was careful to preserve their behavior.

TODO:
- [x] test move command in sim
- [x] replace all prints/fprints with send_feedback
-  ~~test other refactored commands:~~ pretty hard to do in SIM, and most will never be used again (new challenges)


scripts/bash_aliases.sh
```diff
@@ -14,9 +14,8 @@ alias nav="cd $CATKIN_DIR/src/NaviGator"
 alias rnav="ros_connect -n ${HOSTNAMES[1]}"
 alias sshnav="ssh navigator@${HOSTNAMES[1]} -Y"
 
-# Missions
-alias navm="rosrun navigator_missions run_mission"
-alias navc="rosrun navigator_missions move_command"
+# Tasks
+alias nmove="runtask Move"
```
RustyBamboo: why not navc/navm?

ironmig: felt like being explicit about aliases is nice. New people would probably forget navc cause it doesn't have to do with move

RustyBamboo: ah okay, make sense

PR #276, opened 2017-11-14
KILL BOARD: only kill on overall True

ironmig:
* as per Davids recommendation


hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
```diff
@@ -181,7 +184,7 @@ def update_hw_kill(self):
         '''
         Raise/Clear hw-kill ROS Alarm is nessesary (any kills on board are engaged)
         '''
-        killed = any([self.board_status[key] for key in self.board_status])
+        killed = self.board_status['OVERALL']
```
RustyBamboo: Fair

PR #277, opened 2017-11-16
KILL: set problem description in kill alarm

ironmig:
So it looks like this:
fuuu


ironmig:
Note: required uf-mil/ros_alarms#35

PR #288, opened 2018-01-19
CAMERAS: add calibration and launch for See3Cam_cu20

ironmig:
Add calibration for our new HDR camera from e-con systems and a launch file to run the camera in ros using the usb_cam ros driver. The usb_cam driver is a little lacking (no dynamic reconfigure and cannot change the special features on the camera like HDR from their defaults), but it will work now until someone forks/creates a driver specifically for this cam. Way to go open standards (cough, cough pointgrey)!


ironmig:
I ran the cam for several hours with no issues... perhaps it was the connection of cable? We'll see how it goes at the test day

PR #299, opened 2018-02-10
CI: only run tests in NaviGator

ironmig:
depends on uf-mil/mil_common#101

Address #299. Should be getting a lot more green checks after this


ironmig:
Went ahead and redid the kill board test because it never worked. Should be much faster and always pass now

ironmig:
LOOK AT THAT PRETTY GREEN CHECK!!!

PR #301, opened 2018-02-14
The 2D Simulator

Grymestone:
A starting point for the 2D simulator. Still could use some polishing, doesn't start rviz on its own, needs proper params, but it launches just about everything else. It works with the controller and bag files. So its a halfway decent approximation of the NaviGator itself.

### DIFF
--- a/simulation/navigator_2Dsim/launch/owl.launch
+++ b/simulation/navigator_2Dsim/launch/owl.launch
@@ -1,8 +1,11 @@
 <launch>
 
-	<param name="inertia" value="[1, 1, 2]" />
-	<param name="drag" value="[.5, 1, 1]" />
+	<arg name="gui" default="False"/>
+	<param name="inertia" value="[2000, 2000, 4000]" />
+	<param name="drag" value="[1000, 2000, 2000]" />
 	<param name="start_pos" value="[29.534912, -82.303642, 0]"/>
+	<param name="gui" value="$(arg gui)"/>
+
 	<!--  <param name="mass" value="544" /> -->
 	
 	<node pkg="navigator_2Dsim" type="sim2d.py" name="sim2d" output="screen" />
--- a/simulation/navigator_2Dsim/launch/owl.launch
+++ b/simulation/navigator_2Dsim/launch/owl.launch
@@ -1,11 +1,14 @@
 <launch>
 
 	<arg name="gui" default="False"/>
+	<arg name="wind" default="None"/>
+	<!-- kg, kg, kg*m^2 -->
 	<param name="inertia" value="[2000, 2000, 4000]" />
+	<!-- Drag coeffecients for x, y, z -->
 	<param name="drag" value="[1000, 2000, 2000]" />
 	<param name="start_pos" value="[29.534912, -82.303642, 0]"/>
 	<param name="gui" value="$(arg gui)"/>
-
+	<param name="wind" value="$(arg wind)"/>
 	<!--  <param name="mass" value="544" /> -->
 	
 	<node pkg="navigator_2Dsim" type="sim2d.py" name="sim2d" output="screen" />
--- a/simulation/navigator_2Dsim/launch/owl.launch
+++ b/simulation/navigator_2Dsim/launch/owl.launch
@@ -1,0 +1,10 @@
+<launch>
+
+	<param name="inertia" value="[1, 1, 2]" />
+	<param name="drag" value="[.5, 1, 1]" />
+	<param name="start_pos" value="[29.534912, -82.303642, 0]"/>
+	<!--  <param name="mass" value="544" /> -->
+	
+	<node pkg="navigator_2Dsim" type="sim2d.py" name="sim2d" output="screen" />
+
+</launch>
--- a/simulation/navigator_2Dsim/sim2d.py
+++ b/simulation/navigator_2Dsim/sim2d.py
@@ -29,32 +29,39 @@ from std_msgs.msg import Int32, String
 
 class Navsim():
 
-    def __init__(self, t=0, pose0=np.array([0, 0, 0]), twist0=np.array([0, 0, 0]), wind=None):
+    def __init__(self, t=0, pose0=np.array([0, 0, 0]), twist0=np.array([0, 0, 0])):
 
         rospy.init_node('2Dsim')
-        self.des_force = np.array([0, 0, 0])
+
+        self.wind = rospy.get_param("wind")
         self.gui = rospy.get_param("gui")
+        initial_lla = ast.literal_eval(rospy.get_param("start_pos"))
+
         self.world_frame_id = "/enu"
         self.body_frame_id = "/base_link"
         self.state = None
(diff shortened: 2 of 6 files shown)


simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        msg = Odometry()
+        msg.header.stamp = rospy.Time.now()
+        msg.header.frame_id = self.world_frame_id
+        msg.child_frame_id = self.body_frame_id
+        msg.pose.pose.position.x, msg.pose.pose.position.y, msg.pose.pose.position.z = pose
+        # print(msg.pose.pose.position.z)
+        quat = trns.quaternion_from_euler(0, 0, pose[2])
+        msg.pose.pose.orientation = mil_tools.numpy_to_quaternion(quat)
+        msg.twist.twist.linear.x, msg.twist.twist.linear.y, msg.twist.twist.linear.z = twist
```
ironmig: ```msg.twist.twist.linear.x, msg.twist.twist.linear.y = twist[0, 1]```, right?

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+        navsim.twists.append(navsim.twist)
+        navsim.times.append(navsim.t)
+        # Increment sim
+        # wrench = controller.compute_wrench(navsim.pose, navsim.twist, goal_pose,
+        # goal_twist)
+        navsim.state_cb()
+        navsim.publish_odom()
+        navsim.wrenches.append(navsim.des_force)
+        navsim.step(dt, navsim.wrenches[-1])
+        poses = np.asarray(navsim.poses)
```
ironmig: Don't keep ever growing list of previous states if we're not graphing... will just eat more ram as it runs more

simulation/navigator_2Dsim/launch/owl.launch
```diff
@@ -0,0 +1,13 @@
+<launch>
+
+	<arg name="gui" default="False"/>
+	<param name="inertia" value="[2000, 2000, 4000]" />
```
ironmig: Should just have a mass and rotational inertia value, as it can't have a different x and y mass. Also it should be documented what units everything is in

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+        # self.last_ecef = gps.ecef_from_latlongheight(*np.radians(initial_lla))
+        self.last_enu = None
+        self.state_sub_max_prd = rospy.Duration(1 / 100)
+        self.last_state_sub_time = rospy.Time.now()
+
+        self.subscriber = rospy.Subscriber(
+            "/wrench/cmd", WrenchStamped, self.wrench_cb)
+        self.odompublisher = rospy.Publisher(
+            "/odom", Odometry, queue_size=1)
+        self.poses = []
```
ironmig: These are here and also down in the main loop???

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+        # Boat is longer than shorter, hence .5, 1, 1.
+        self.drag = np.array(ast.literal_eval(
+            rospy.get_param('drag')), dtype=np.float64)
+        # Time in seconds
+        self.t = t
+        # Wind is a function of time that returns a wrench
+        if wind:
+            self.wind = wind
+        else:
+            self.wind = lambda t: np.zeros(3, dtype=np.float64)
```
ironmig: Wind should be configurable or just removed

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+
+        rospy.Timer(rospy.Duration(1 / 100), self.publish_odom)
```
ironmig: Wouldn't it make more sense to just have one timer that updates state and publishes odom instead of two?

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def state_deriv(self, wrench, R):
+        posedot = R.dot(self.twist)
+        twistdot = (1 / self.inertia) * (wrench - self.drag *
+                                         self.twist + R.T.dot(self.wind(self.t)))
+        return posedot, twistdot
+
+    def wrench_cb(self, msg):
+        ''' Grab new wrench
+                This is the 'b' in Ax = b
```
ironmig: not really a relevant comment  in this context

simulation/navigator_2Dsim/sim2d.py
Grymestone: Ah good point, I'll toss it into the if statement.

simulation/navigator_2Dsim/launch/owl.launch
Grymestone: Units are in the main python file but it does make more sense to put them in the launch file where people will actually be changing them.

simulation/navigator_2Dsim/sim2d.py
Grymestone: I'll make it an arg with the default value None.

Grymestone:
As a note, had to add a one time rospy.sleep() to the main due to the fact it hit the publish command before it could be spun up to publish. This was likely caused by removing the list creations I put at the front of it.

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,264 @@
...
+    dt = .05
+    rospy.sleep(5)
+
+    while not rospy.is_shutdown():
+
+        # Increment sim
+        navsim.state_cb()
+        navsim.publish_odom()
+        navsim.wrenches.append(navsim.des_force)
+        navsim.step(dt, navsim.wrenches[-1])
```
ironmig: Looks like this will just run in an infinite loop with no sleep now... Again I recommend doing all state calculations in a ROS timer. Nothing needs to be or should be dependent on unpredictable timing events happening at the right time.

simulation/navigator_2Dsim/sim2d.py
```diff
@@ -0,0 +1,264 @@
...
+    # print(navsim.gui)
+    dt = .05
+    rospy.sleep(5)
+
+    while not rospy.is_shutdown():
+
+        # Increment sim
+        navsim.state_cb()
+        navsim.publish_odom()
+        navsim.wrenches.append(navsim.des_force)
```
ironmig: Still keeps infinite list!

simulation/navigator_2Dsim/launch/owl.launch
ironmig: See first part of comment

ironmig:
So I need to use this tonight to test missions, so I'm gonna fix it up and you can checkout the diff

ironmig:
Included in #307

PR #304, opened 2018-02-16
KILL BOARD: add controller message handling functionality

zobeld:
- Add controller message identifier to the byte handling loop

- Create publisher to /joy_emergency for received joy data, handled by navigator_emergency.py

- Add some constants for the controller messages


zobeld:
- Added navigator_emergency.py to gnc launch file

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
```diff
@@ -81,12 +97,51 @@ def run(self):
...
+            self.ctrl_msg_timeout = rospy.Time.now()
+            return
+        # If receiving the controller message, record the byte as stick/button data
+        if (self.ctrl_msg_count > 0) and (self.ctrl_msg_count <= 8):
+            # If 1 second has passed since the message began, timeout and report warning
+            if (rospy.Time.now() - self.ctrl_msg_timeout) >= rospy.Duration(1):
+                self.ctrl_msg_received = False
+                self.ctrl_msg_count = 0
+                rospy.logwarn('Timeout receiving controller message. Please disconnect controller.')
+            if self.ctrl_msg_count > 2:  # The first 6 bytes in the message are stick data bytes
```
ironmig: If it's the first 6 bytes, why is this >2? Is it actually the last 6 bytes?

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
```diff
@@ -81,12 +97,51 @@ def run(self):
...
+                else:  # Odd number byte: combine two bytes into a stick's data word
+                    self.sticks_temp += int(msg.encode("hex"), 16)
+                    if (self.ctrl_msg_count > 6):
+                        self.sticks['UD'] = self.sticks_temp
+                    elif (self.ctrl_msg_count > 4):
+                        self.sticks['LR'] = self.sticks_temp
+                    else:
+                        self.sticks['TQ'] = self.sticks_temp
+                    self.sticks_temp = 0x0000
+            else:  # The last 2 bytes are button data bytes
```
ironmig: You mean first 2?

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
zobeld: When the controller message start byte (\xA0) is received, the ctrl_msg_count is set to 8 to indicate how many bytes are remaining, and decreases. If the count is more than 2, it's not the last 2 bytes yet.

hardware_drivers/navigator_kill_board/nodes/kill_board_driver.py
zobeld: As described above, these are the last 2. This logically works out, but I see how this can cause confusion. I'll rewrite it to be more readable, with the count increasing as it reads bytes.

PR #307, opened 2018-02-18
SIMULATION: add simple 2d kinematic simulator for GNC and high level testing

ironmig:
Build on #301, getting the 2D sim to a functional state and cleaning up the code so it is more readable and contains less dead code.


mission_control/navigator_launch/launch/sim2d.launch
```diff
@@ -0,0 +1,25 @@
+<launch>
+    <!-- Run 2D Simulator -->
+    <include file="$(find navigator_2dsim)/launch/owl.launch" />
+
+    <param name="/is_simulation" type="bool" value="True" />
+
+    <!-- Launch Fake Shooter Service Provider -->
+    <node pkg="navigator_shooter" type="ShooterControl.py" name="shooter_controller">
```
Grymestone: I copied this in from other launch files, is it strictly necessary here? Also aren't we going to have to rewrite this with Kip's new launcher?

Grymestone: If this is nothing then its good to merge.

ironmig: Good catch, I don't think anything uses that functionality so I'll remove it from the launch files in sim and IRL.

Issue #310, opened 2018-02-23
2D SIM: refine kinematic constants


ironmig:
Currently, the kinematic constants in the 2D sim are fairly random magic numbers for mass, drag coefficients, and rotational inertia. These values should be set to something close to their true values using bag data, our knowledge of the boat, etc.

ironmig:
* [x] Mass
* [x] Drag
* [ ] Rotational Inertia

ironmig:
2D is no longer used much, so closing

PR #311, opened 2018-03-01
2D SIM: update drag coefficents from bag data

ironmig:
#310


ironmig:
#Non-arbitrary-Units

RustyBamboo:
mmmm capybaras

PR #315, opened 2018-03-12
KILL BOARD: replace remote constant with heartbeats and button constants

zobeld:
- Rename 'REMOTE' kill constant with 'HEARTBEAT_COMPUTER' constant

- Add 'BUTTON_REMOTE' and 'HEARTBEAT_REMOTE' constants to reflect changes to data sent by kill board microprocessor

- These changes don't alter the actual kill functionality, only report more accurately the source of the kill


ironmig:
Worked at test day 03/18.

PR #328, opened 2018-03-20
ALIAS: add navtest alias

ironmig:
* ALIAS: add navtest alias
* CI: format check all python files

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,18 +15,21 @@ dockerNode(image: 'uf-mil:navigator') {
 		'''
 	}
 	stage("Test") {
-		sh '''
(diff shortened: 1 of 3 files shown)


RustyBamboo:
Looks like it works but where is
```
Running shell script

bash: cannot set terminal process group (8638): Inappropriate ioctl for device

bash: no job control in this shell

The master roscore is set to http://localhost:11311
```
coming from?

ironmig:
Yeah that is annoying... I had to add run the Jenkins commands with the env string ```#!/bin/bash -i``` to make it an interactive terminal so that the commands can call the aliases... but it does print that error. However, this is not what is causing the build to fail, it is a pep8 error

PR #330, opened 2018-03-23
2D SIM: fix drag to be proportional to velocity^2

ironmig:
not velocity


ironmig:
After talking to Jason, it isn't super clear that this is better, but w/e. This PR also solves a bug where position.z is our angle

Issue #323, opened 2018-03-18
Pipe sdgps to null


ironmig:
AFAIK we don't use the logs from the sdgps software. It has caused issues by consuming hard drive write speed, didn't space, etc so we should just pipe in to null

PR #332, opened 2018-03-23
BOOT, CAMERAS, MOVE: pre lake day omnibus

ironmig:
Will run syphase gps pipeline in just one command. We no longer need the raw data logged (and can't afford the hard drive space / bandwidth). 

#323 

**sdgps is no longer run under sudo, this requires some permissions changes we must implement on NaviGator (Kevin has the instructions)**


ironmig:
TOO MUCH POOOOOWEEEERR!

Issue #351, opened 2018-08-21
Integrate VMRC Simulator


ironmig:
Replace parts of navigator's gazebo simulator configuration/models/plugins with those found in [VMRC](https://bitbucket.org/osrf/vmrc/src/default/), the official virtual competition of RobotX. This will provide more realistic boat physics. a more accurate course, and better performance.

* [ ] Use VMRC's WAM-V model
* [ ] Use VMRC's sandisland model and competition element models
* [ ] Use VMRC's gazebo plugins for dynamics, wind, and waves
* [ ] Write/Convert the gazebo thruster plugin to provide the same interface as the roboteq motor drivers

PR #355, opened 2018-09-03
SIMULATION: integrate VMRC to replace old gazebo simulation

ironmig:
connects to #351
connects to #326

[Video of simulation](https://drive.google.com/file/d/1tBpMdTSftdurNBDZbAxJhxEApsntcHmh/view)

Omnibus PR that refactors navigator_gazebo to use VMRC (the software for running the simulated version of RobotX), which provides a more accurate visual simulation as well as things like dynamics, wind, and waves.

With this, Navigator is simulated down to individual forces applied by the motors. A plugin has been created to simulate the roboteq motor drivers

Some other packages changes
* Add a git-ifed version of VMRC as a submodule
* Update NaviGator URDF model to have accurate sensor positions.
* Use robot_state_publisher to publish tf based on URDF, so frames poses are not duplicated
* Depcreate all the old models/plugins that VMRC now provides

Currently missing (future work):
* Velodyne simulation
* Feedback from simulated motor drivers (needed for testing battery-low failures)


Grymestone:
CI needs to be updated or nothing will build after this is merged.

PR #363, opened 2018-09-06
ALIASES: add bag-fast alias to bag to SSD

### LINKED ISSUE #321: Solve bagging buffer issues
We currently bag to an external USB3 (spinning disk?) hard drive. It has a lot more storage than the SSD but is much slower. Sometimes when bagging high bandwidth topics like cameras, the write buffer is exceeded and the bag is corrupted. We need some utility / program that will allow us to bag initially to the SSD, but copy it over to the hard drive afterwards, finally deleting it from the SSD to recover the space.

ironmig:
closes #321 

Adds an alias bag-fast to bag to a directory on the SSD which has sufficient write speed to handle high bandwidth topics. Then use aliases syncbags and purgebagfast to sync the SSD bags to the external HDD and clear the ones on the SSD.


Grymestone:
What would be nice in the future is the script also publishing the amount of memory left after each bag is finished recording.

PR #378, opened 2018-09-20
DATASETS: Add 6 new datasets to LabelBox

Grymestone:


perception/navigator_vision/datasets/project_bags.yaml
```diff
@@ -213,3 +213,571 @@ datasets:
...
+          file: 2018-02-04/blackbuoy.bag
+          topic: /camera/front/left/image_raw    
+          encoding: rect_color              
+          freq: 0.15           
+          start: 0.0
+          stop: 36.0  
+          # End of 2018-02-04 Folder
+          # End of point gray bags
+    - 
+      name: navigator_seecam_pg-2018-09-09
```
ironmig: seecam_pg? should contain camera and location

perception/navigator_vision/datasets/project_bags.yaml
```diff
@@ -213,3 +213,571 @@ datasets:
         start: 0
         stop: 9
 
+  -
+      name: navigator_wahlberg_pg_2018-02-04
```
ironmig: Please follow convention from first dataset of `pointgrey` rather than `pg`, so we can later segment it by camera if needed

PR #383, opened 2018-09-23
navigator_robotx_comms package

### LINKED ISSUE #358: Write interface for RobotX Communication Protocol
RobotX 2018 requires teams to communicate with the judge's computer over a specific protocol [detailed here](https://www.robotx.org/images/RobotX-2018-Communications-Protocol_v0.5.pdf).

* [ ] Write a ROS node ```navigator_robotx_comms``` to send the heartbeat message
* [ ] Add functionality to the above node to send the specialized messages specific to certain challenges, providing a ROS service to send this message
* [ ] Write a test implementing the "server" side of the protocol to test the above node.
* [ ] Wherever possible, implement the protocol as a library so functionality can be reused in the server/client/test executables

BSpwr:
closes #358

Initial heartbeat function along with testing server for implementing RobotX communication protocol.


utils/navigator_robotx_comms/test/set_localhost_params.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
```
ironmig: Instead of a script to set params, you can set them in the [launch/.test file](http://wiki.ros.org/roslaunch/XML/param) or [runtime](http://wiki.ros.org/Parameter%20Server#Private_Parameters)

utils/navigator_robotx_comms/package.xml
```diff
@@ -0,0 +1,68 @@
...
+  <!-- Dependencies can be catkin packages or system dependencies -->
+  <!-- Examples: -->
+  <!-- Use depend as a shortcut for packages that are both build and exec dependencies -->
+  <!--   <depend>roscpp</depend> -->
+  <!--   Note that this is equivalent to the following: -->
+  <!--   <build_depend>roscpp</build_depend> -->
+  <!--   <exec_depend>roscpp</exec_depend> -->
+  <!-- Use build_depend for packages you need at compile time: -->
+  <!--   <build_depend>message_generation</build_depend> -->
+  <!-- Use build_export_depend for packages you need in order to build against this package: -->
```
ironmig: Good job using ```catkin_create_pkg```... saves a lot of time. However, we remove the commented out lines so it's easy to see what is actually running in package.xml and CMakeLists.txt

utils/navigator_robotx_comms/nodes/navigator_robotx_comms.py
```diff
@@ -0,0 +1,157 @@
...
+            quaternion = self.odom.pose.pose.orientation
+            north_south = ""
+            east_west = ""
+            if quaternion.x > 0:
+                east_west = "E"
+            elif quaternion.x < 0:
+                east_west = "W"
+            if quaternion.y > 0:
+                north_south = "N"
+            elif quaternion.y < 0:
```
ironmig: We talked about this at lakeday, but remember to update this so use the yaw angle, which is much easier to reason about for something like this cardinal direction identification

utils/navigator_robotx_comms/nodes/navigator_robotx_comms.py
```diff
@@ -0,0 +1,157 @@
...
+            socketConnection = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            # attempt to reconnect, otherwise sleep for 2 seconds
+            try:
+                socketConnection.connect((TCP_IP, TCP_PORT))
+                connected = True
+                print 'Connection to TD Server Successful'
+            except socket.error:
+                time.sleep(2)
+        heartbeatMsg = robot_x_comms.heartbeat_msg()
+        while True:
```
ironmig: We avoid ```while True``` as it can sometimes make your program impossible to close. Consider using ```while not rospy.is_shutdown()```, or even better using rospy.Timer to have ROS handle this regular interval function call for you

utils/navigator_msgs/srv/MessageHeartbeat.srv
```diff
@@ -0,0 +1,2 @@
+---
+string message
```
ironmig: Is this used? Heartbeat should be sent automatically at fixed rate, not triggered externally with a service

utils/navigator_msgs/srv/MessageDetectDeliver.srv
```diff
@@ -0,0 +1,4 @@
+string shape_color
+string shape
+---
```
ironmig: Provide comments for what are valid values for these fields. Do I put "CROSS" or "cross", "RED" or "r"?

utils/navigator_robotx_comms/nodes/navigator_robotx_comms.py
```diff
@@ -0,0 +1,413 @@
...
+        self.team_id = rospy.get_param('team_id')
+        self.message_id = "RXHRB"
+        self.gps_array = None
+        self.odom = None
+        self.auv_status = 1
+        self.system_mode = 2
+
+        rospy.Subscriber("lla", PointStamped, self.gps_coord_callback)
+        rospy.Subscriber("odom", Odometry, self.gps_odom_callback)
+        # rospy.Subscriber("auv_status", String, self.auv_status_callback)
```
ironmig: As this is the class for the message, it should not contain subscribers, just serialize/serialize the message. Put these in the client class

utils/navigator_robotx_comms/nodes/navigator_robotx_comms.py
```diff
@@ -0,0 +1,413 @@
...
+            try:
+                self.socket_connection.send(message)
+                break
+            except socket.error:
+                print('Connection to TD Server Lost, Attempting Reconnection')
+                self.connected = False
+                # recreate socket
+                self.recreate_socket()
+                while not self.connected and not rospy.is_shutdown():
+                    # attempt to reconnect, otherwise sleep for 2 seconds
```
ironmig: Looks like duplicate of the connect() function, so why not just call it?

utils/navigator_robotx_comms/nodes/navigator_robotx_comms.py
```diff
@@ -0,0 +1,413 @@
...
+        if not self.connected:
+            print('Attempting Connection to TD Server')
+        while not self.connected and not rospy.is_shutdown():
+            # recreate socket
+            self.socket_connection = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            # attempt to reconnect, otherwise sleep for 2 seconds
+            try:
+                self.recreate_socket()
+                self.connected = True
+                print('Connection to TD Server Successful')
```
ironmig: use `rospy.loginfo` so the prints are seen in the launch terminal and written to file

utils/navigator_robotx_comms/test/robotx_comms.test
```diff
@@ -0,0 +1,8 @@
+<launch>
+    <test test-name="robotx_comms_test" pkg="navigator_robotx_comms" type="robotx_comms_test.py" />
+</launch>
+
+	<test test-name="set_localhost_params" pkg="navigator_robotx_comms" type="set_localhost_params.py" />
```
ironmig: This is still not needed, just set the params in the launch file, like:
```
 <node name="navigator_robotx_comms" pkg="navigator_robotx_comms" type="navigator_robotx_comms.py" >
   <param name="~my_param" value="foo" />
</node>
```

ironmig: Note that the `~` makes it a "private" parameter, so in the node it should be accessed as `rospy.get_param("~my_param")

PR #386, opened 2018-09-25
BOOT: run sonar ros_bridge on boot

ironmig:
* We cant put it in normal launch files b/c its proprietary software


ironmig:
It is closed source, proprietary software so we cannot put it in this repo. It is also not in a catkin package, so it cannot be found in a launch file though I supose we could hard code it in. Currently the program does not respond to SIGINT, so having it in our core launch system would likely freeze attempts to relaunch / shutdown

PR #399, opened 2018-10-09
CAMERAS: fix front_left serial id and disable it

ironmig:
At lakeday we discovered front_left has the wrong ID. Once it was fixed, we couldn't run front_left and starboard at same time (thank pointgrey), so disable front_left for now.


ironmig:
Underlying issue ticketed in #407

PR #400, opened 2018-10-12
[WIP] Start Gate mission

jaxnb:
depends on #381 

Very WIP
This is the first revision of the start gate mission. It will integrate with Marshall's buoy mission.

Known issues:
- Needs to differentiate between entering and exiting
- Hydrophones hardcoded - ish.
- Needs real documentation
- Kill needs to be adjusted, pending #402


jaxnb:
Replaced with #413

PR #401, opened 2018-10-14
Start gate marshall

MarshallRawson:
depends on #381

Mission to run in pcodar branch.
executes the going around of the bouy for the qualifier challenge.

currently uses multiple individual way points to go around the bouy and stops at each way point

should probably utilize the circle function, but this works


ironmig:
Looks like CI is failing on an import line. You will also need to format your files to pass the linter. You can use autopep8 to do some of the work automatically. you can verify the formatting locally with the `navfmt` alias

PR #414, opened 2018-11-02
[WIP] Machine Learning Classifier for NaviGator

Grymestone:
Depends on [#138](https://github.com/uf-mil/mil_common/pull/138).
Very much a work in progress, this is basically a straight port from SubjuGator code. There are quite a few things that could improve this and cut down on file usage but as it stands this code functions and will return a classified image with bounding box. 

TODO

- [ ] Move all Utils and Protocs to a single directory
- [ ] Have those Utils and Protocs take in arguments rather than being hard coded to look at a single file
- [ ] Modify the perception scripts to return what we need for each task beyond just the bounding box.
- [ ] Integrate with PCODAR. 

Currently I have another branch that was previously integrated with PCODAR before its completion. I'm not sure how well it will work now but I have a baseline at least for the integration.


ironmig:
I think it is reasonable to target this for mil_common.

The goal is to have a node that implements the [vision_node interface](https://github.com/uf-mil/mil_common/blob/master/perception/mil_vision/mil_vision_tools/vision_node.py) and takes in the necessary configuration files (graph, label map) as parameters. All this nodes has to do is implement a callback function which takes in an image and returns a list of labeled boxes. It is a class, so you can override __init__ for startup configuration (loading in the graph). 

As for the large graph files, I recommend hosting them on plumbusi and using [file(DOWNLOAD)](https://cmake.org/cmake/help/latest/command/file.html#download) in CMake to have it be downloaded on builds (verified with checksum) without having to have it as a git object.

ironmig:
Another question: Is killing/restarting the processes the only solution for tensorflow's restrictions on multiple processes? For example, maybe you just need to close/restart the tensorflow session object? Worth a small amount of research to have a cleaner/faster solution.

PR #415, opened 2018-11-03
Controller

zobeld:
KILL BOARD: update emergency controller functionality

- Update emergency controller node to have raise/clear kill buttons (no longer toggle). Remove autonomous, wrench toggle, and shooter functionality. Remove clear kill from controller startup.

- Update kill board constants for similar functionality to xbox controller, and change interpretation of "button bytes" from incoming messages from emergency controller to reflect updated functions.

- Update kill board node to match new constants and functionality. Change interpretation of incoming message joystick bytes such that the software's 'x axis' corresponds to joystick forward/back and 'y axis' corresponds to joystick left/right.

### DIFF
--- a/utils/remote_control/navigator_emergency_control/nodes/navigator_emergency.py
+++ b/utils/remote_control/navigator_emergency_control/nodes/navigator_emergency.py
@@ -32,17 +32,17 @@ class Joystick(object):
         Used to reset the state of the controller. Sometimes when it
         disconnects then comes back online, the settings are all out of whack.
         '''
-        self.last_kill = False
+        self.last_raise_kill = False
+        self.last_clear_kill = False
         self.last_station_hold_state = False
-        self.last_auto_control = False
         self.last_emergency_control = False
-        self.last_shooter_cancel = False
-        self.last_change_mode = False
+        self.last_go_inactive = False
+        self.thruster_deploy_count = 0
+        self.thruster_retract_count = 0
 
         self.start_count = 0
         self.last_joy = None
         self.active = False
-
         self.remote.clear_wrench()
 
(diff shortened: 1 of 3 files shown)


utils/remote_control/navigator_emergency_control/nodes/navigator_emergency.py
```diff
@@ -70,48 +70,62 @@ def joy_recieved(self, joy):
...
+            self.thruster_deploy_count += 1
+        else:
+            self.thruster_deploy_count = 0
+
+        if self.thruster_retract_count > 10:
+            self.remote.retract_thrusters()
+            self.thruster_retract_count = 0
+        elif self.thruster_deploy_count > 10:
+            self.remote.deploy_thrusters()
+            self.thruster_deploy_count = 0
```
ironmig: This should go before the guard on 109 ignoring commands when inactive

utils/remote_control/navigator_emergency_control/nodes/navigator_emergency.py
```diff
@@ -70,48 +70,62 @@ def joy_recieved(self, joy):
...
-        shooter_cancel = bool(joy.buttons[1])
-        change_mode = bool(joy.buttons[3])
+        emergency_control = bool(joy.buttons[13])
+        thruster_retract = bool(joy.buttons[4])
+        thruster_deploy = bool(joy.buttons[5])
+
+        if go_inactive and not self.last_go_inactive:
+            rospy.loginfo('Back pressed. Going inactive')
+            self.reset()
+        return
```
ironmig: Looks like this will just always return based on indention

PR #419, opened 2018-11-04
LAUNCH: run sick with hardware.launch

ironmig:


ironmig:
Seems like SICK will not be useful after all

PR #421, opened 2018-11-09
Nav_tui Review

### LINKED ISSUE #289: Create curses based status/debug TUI
We have a nice Qt based status GUI which displays things like battery voltage, autonomous/remote control mode, etc. However, much of the time at test days and competition is spent in a terminal working on code so we rarely look at this GUI. Curses is a TUI (terminal user interface) library which we use for programs like [ros alarms monitor](https://github.com/uf-mil/ros_alarms/blob/master/nodes/monitor) to create terminal based graphical interfaces.  It would be great to have a TUI for the same sorts of status items we have in the Qt GUI:

Some basic features
* Battery Voltage (color depends on battery low / critical / okay)
* Wrench mode (autonomous or remote controlled)
* kill status
* GPS coordinates???? that would look hella pro

ericgorday:
closes #289 

NaviGator diagnostics TUI
Currently Displays: wrench mode, kill status, battery voltage, LLA coordinates, and individual thruster faults (if any)
Issue: #289

Possible concerns:
-I used the try-except block from the voltage gui that attempts to get the low-voltage and critical-voltage parameters from the parameter server. It uses 26 and 20 as the low and critical voltage threshold ,respectively, if the parameters haven't been set.
-Wrench mode displays: '/wrench/selected' in the Gazebo simulator initially before movement commands are given for the navigator 
-Values given for altitude look strange but the values of latitude and longitude match the simulator


nav_tui.py
```diff
@@ -0,0 +1,216 @@
...
+		self.BL_fault = 0
+		self.BR_fault = 0
+		self.motor_topics = ['/FL_motor', '/FR_motor', '/BL_motor', '/BR_motor']
+		self.fault_codes = {1: 'OVERHEAT', 2: 'OVERVOLTAGE', 4: 'UNDERVOLTAGE', 8: 'SHORT_CIRCUIT',
+                            16: 'EMERGENCY_STOP', 32: 'SEPEX_EXCITATION_FAULT', 64: 'MOSFET_FAILURE',
+                            128: 'STARTUP_CONFIG_FAULT'}
+
+		#Will try and access low/critical voltage ros parameters, if none are located default values are used
+		try: 
+            		self.lowThreshold = rospy.get_param('battery-voltage/low')
```
ironmig: `self.lowThreshold = rospy.get_param('battery-voltage/low', default=26)` works

ericgorday:
The multiple commits were me just fumbling around with the fork-commit systems on github, it's my first pull request

nav_tui.py
```diff
@@ -0,0 +1,216 @@
...
+            		self.criticalThreshold = rospy.get_param('battery-voltage/critical')
+        	except:
+            		print("Low and Critical Voltage not set, Low: 26 Critical:20")
```
ironmig: Always [roslog](http://wiki.ros.org/rospy/Overview/Logging) instead of printing

nav_tui.py
ericgorday: I have fixed this in my most recent commit, thanks.

nav_tui.py
ericgorday: Thanks, will keep this in mind for the future.

nav_tui.py
```diff
@@ -0,0 +1,229 @@
...
+                            16: 'EMERGENCY_STOP', 32: 'SEPEX_EXCITATION_FAULT', 64: 'MOSFET_FAILURE',
+                            128: 'STARTUP_CONFIG_FAULT'}
+
+        # Will try and access low/critical voltage ros parameters, if none are located default values are used
+        self.lowThreshold = rospy.get_param('battery-voltage/low', default=26)
+        self.criticalThreshold = rospy.get_param('battery-voltage/critical', default=20)
+
+    # Cb function for wrench type
+    def wrench_cb(self, string):
+        self.wrench = string.data
```
ironmig: Helpful to trim the `/wrench` from this topic if it exists, so:
`/wrench/autonomous` -> `autonomous`
`autonomous` -> `autonomous`

nav_tui.py
```diff
@@ -0,0 +1,229 @@
...
+        self.window.clear()
+        # Initialization of color pairs for text display
+        color = curses.A_NORMAL
+        curses.init_pair(1, curses.COLOR_RED, curses.COLOR_BLACK)
+        curses.init_pair(2, curses.COLOR_YELLOW, curses.COLOR_BLACK)
+        curses.init_pair(3, curses.COLOR_GREEN, curses.COLOR_BLACK)
+        curses.init_pair(4, curses.COLOR_BLACK, curses.COLOR_WHITE)
+        curses.init_pair(5, curses.COLOR_YELLOW, curses.COLOR_BLACK)
+
+        while True:
```
ironmig: An infinite while loop with no wait is problematic. My screen flickers running this and it hogs a lot of cpu. Consider updating at a fixed rate (like 1 hz)

ironmig:
Also please:
* [Make the file executable](http://lmgtfy.com/?q=how+to+make+file+executable+in+linux)
* Rename to `uilts/navigator_tools/nodes/navigator_status_tui` (no .py is okay)

nav_tui.py
ericgorday: Yeah I hadn't before noticed the cpu hit of running the tui but it's significant. I set the rate to be 2 hz and it greatly reduces the cpu load.

nav_tui.py
ericgorday: I added an if-condition for this. If it returns true that the string message contains: "/wrench/" and trims that part off, otherwise it just returns the string.

ironmig:
File is still not executable, refer to my previous comment about how to verify that it is

ironmig:
Your last commit has no changes. If you change permissions on a file in git, this change must be commited. `chmod +x <file>` 
`git commit <file>`

ironmig:
You're right it can be run with rosrun (I guess rosrun knows to run it in a python interpreter), but it is not executable so won't tab complete and can't be run from within the directory with `./navigator_status_tui`. Perhaps you're developing on windows or on a legacy filesystem?

Either way, just save this to a file (my_patch.txt), then apply it with `git apply <my_patch.txt>`

`
diff --git a/utils/navigator_tools/nodes/navigator_status_tui b/utils/navigator_tools/nodes/navigator_status_tui
old mode 100644
new mode 100755
`

PR #448, opened 2018-11-28
Entrance Gate Mission

jaxnb:


jaxnb:
Depends on https://github.com/uf-mil/mil_common/pull/163

ironmig:
Looks like CI is upset about some formatting issues. Props for noting dependencies, if you want them to appear correctly on waffle, you must [follow a certain format and the text must be on the PR description](https://help.waffle.io/dependencies/which-keywords-are-supported-with-dependencies), which you can edit

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+        return [pvec[0] / r, pvec[1] / r, 0]
+
+    '''
+
+        Perception Utilities
+
+    '''
+
+    @util.cancellableInlineCallbacks
+    def find_nearest_objects(self, name, number):
```
ironmig: To maximize code reuse, perhaps you could instead use the function added in #443

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+            hydrophones_to_enu = yield self.tf_listener.get_transform('enu', p_message.header.frame_id)
+            hydrophones_origin = hydrophones_to_enu._p[0:2]
+            heading = rosmsg_to_numpy(p_message.vector)
+            heading_enu = hydrophones_to_enu.transform_vector(heading)
+            heading_enu = heading_enu[0:2] / np.linalg.norm(heading_enu[0:2])
+
+            # Track the ping
+            self.intersect_vectors.append((hydrophones_origin[0:2], hydrophones_origin[0:2] + heading_enu[0:2]))
+        except tf2_ros.TransformException, e:
+            rospy.logwarn('TF Exception: {}'.format(e))
```
ironmig: can't rospy in txros program

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+from txros import util
+from navigator_missions.navigator import Navigator
+import numpy as np
+from mil_tools import rosmsg_to_numpy
+from twisted.internet import defer
+import math
+from mil_misc_tools import ThrowingArgumentParser
+from argparse import ArgumentTypeError
+import tf2_ros
+import rospy
```
ironmig: can't rospy in txros program

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+                self.end = end
+
+            def __eq__(self, other):
+                return self.start <= other <= self.end
+
+        parser = ThrowingArgumentParser(description='Start Gate Mission',
+                                        usage='''Default parameters: \'runtask EntranceGate --passscan false
+                                         --multilateration false --scandist 10 --speed 0.75 --kill true
+                                         --scantime 10 --traversaldist 7\'''')
+        parser.add_argument('-p', '--passscan', type=str2bool, default=False,
```
ironmig: More conventional to use ```type='set_true'``` like [here](https://github.com/uf-mil/NaviGator/blob/c3408b30b5ad4a62211f4fb056efc0b9be738519/mission_control/navigator_missions/navigator_missions/move.py#L32). So you would just do `... -m` for multiliteration rather than `... -m true`

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+            yield self.hydrophones.disable()
+            yield self.hydrophones.reset()
+        else:
+            # Reset pings
+            self.intersect_vectors = []
+
+        # Execute Scan 1
+
+        self.send_feedback('Navigating to scan start point 1')
+        yield self.move.set_position(scan_point_0).look_at(scan_lookat_0).go()
```
ironmig: This should probably be .look_at().set_position() to minimize travel time. Looking at a point you are already on top of is probably unreliable / random.

mission_control/navigator_missions/navigator_missions/entrance_gate.py
```diff
@@ -0,0 +1,463 @@
...
+            self.hydrophones.set_callback(None)
+
+        # If kill is enabled, revive the thrusters
+        if should_kill:
+            self.enable_autonomous()
+
+        # Execute Scan 2
+
+        self.send_feedback('Navigating to scan start point 2')
+        yield self.move.set_position(scan_point_1).look_at(scan_lookat_1).go()
```
ironmig: Same as above

jaxnb:
The two modes are, in fact, different.
The two points mode moves the boat to two positions, at each, it will either station keep or drift. This gives us two different areas that we could scan from and gives us the opportunity to kill the boat. This is how we've been testing it.
The pass mode causes the boat to make one long pass in front of the gates, which means there may be a more even distribution set of origins for the lobs. We tested this on the last lake day and found that it seemed to be effective.

We will have to figure out which mechanism we want to use through actual testing on the water.

mission_control/navigator_missions/navigator_missions/entrance_gate.py
jaxnb: Fixed https://github.com/jaxnb/NaviGator/commit/e26de47ddb4335edd1105f6102eb5d21daf7e730

mission_control/navigator_missions/navigator_missions/entrance_gate.py
jaxnb: Fixed https://github.com/jaxnb/NaviGator/commit/e26de47ddb4335edd1105f6102eb5d21daf7e730

mission_control/navigator_missions/navigator_missions/entrance_gate.py
jaxnb: Fixed https://github.com/jaxnb/NaviGator/commit/72279f71d82bbfb332a10b010744dba891e938bc

mission_control/navigator_missions/navigator_missions/entrance_gate.py
jaxnb: It's actually looking at a point that's on the other side of the gate, perpedicular through the gate. 'get_perpendicular_points' returns two points, one on the boat's side of the gate, and the other on the opposite side of the gate.
https://github.com/jaxnb/NaviGator/blob/startgate/mission_control/navigator_missions/navigator_missions/entrance_gate.py#L95

mission_control/navigator_missions/navigator_missions/entrance_gate.py
jaxnb: See above

PR #486, opened 2018-12-14
Scan the code mission

jaxnb:


mission_control/navigator_missions/navigator_missions/stc_jaxon.py
```diff
@@ -0,0 +1,39 @@
...
+        yield self.move.look_at(self.stc).go()
+
+        yield self.move.set_position(self.stc).backward(5).yaw_left(1.57).go()
+
+        # Get scan the code stuff
+        stc_result = yield self.stcsub.get_next_message()
+        stc_result = stc_result.color_pattern
+        self.net_stc_result = stc_result
+
+        if stc_result == 'R':
```
ironmig: stc_result will be 3 chars, like RGB. Perhaps this should be stc_result[0]?

